PR #1, opened 2018-07-23
feat: transaction module

NicoSantangelo:


src/modules/transaction/middleware.ts
```diff
@@ -0,0 +1,24 @@
+import { fetchTransactionRequest } from './actions'
+import {
+  isTransactionAction,
+  getTransactionHashFromAction
+} from './utils'
+import { getAddress } from '../wallet/selectors'
+import { RootMiddleware } from '../../types'
+
+export const createTransactionMiddleware = () => {
```
cazala: Do we need this to be a function? Couldn't we just export `transactionMiddleware` as is?

cazala:
🏅 !

src/modules/transaction/middleware.ts
NicoSantangelo: Yes! I don't think that having multiple middlewares makes sense, but: I wanted to mantain

NicoSantangelo: Yes! I don't think that having multiple middlewares makes sense, but:
1) I wanted to maintain the "convention" of creating a new middleware on the store, like `routerMiddleware` or createSagasMiddleware` do
2) I think it might be useful to keep it this way in case that we need to add a parameter in the future

what do you think?

cazala: agreed 👍

PR #11, opened 2018-09-13
BREAKING CHANGE: storage migration

### LINKED ISSUE #10: Versioning for localStorage keys

nachomazzara:
close #10 

Add migrations for localstorage keys

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ function* handleFetchLandAmountRequest(action: FetchLandAmountRequestAction) {
 ## Storage
 
 The storage module allows you to save parts of the redux store in localStorage to make them persistent and migrate it from different versions without loosing it.
-This module is required to use other modules like `Transaction`, `Translation`, `Wallet` and `Storage`.
+This module is required to use other modules like `Transaction`, `Translation` and `Wallet`.
 
 ### Installation
 
@@ -314,7 +314,7 @@ const { storageMiddleware, loadStorageMiddleware } = createStorageMiddleware({
   storageKey: 'storage-key' // this is the key used to save the state in localStorage (required)
   paths: [] // array of paths from state to be persisted (optional)
   actions: [] // array of actions types that will trigger a SAVE (optional)
-  migrations: migrations // migration object that will migrate your localstorage
+  migrations: migrations // migration object that will migrate your localstorage (optional)
 })
 
 const middleware = applyMiddleware(
@@ -335,7 +335,6 @@ loadStorageMiddleware(store)
 
 ```ts
 export const migrations = {
-  1: migrateToVersion1(data),
   2: migrateToVersion2(data),
   3: migrateToVersion3(data)
 }
@@ -344,7 +343,7 @@ export const migrations = {
 Where every `key` represent a migration and every `method` should return the new localstorage data:
 
 ```ts
-function migrateToVersion1(data) {
+function migrateToVersion2(data) {
   return omit(data, 'translations')
 }
 ```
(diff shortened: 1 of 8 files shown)


src/lib/localStorage.ts
```diff
@@ -11,6 +13,34 @@ export function hasLocalStorage(): boolean {
...
+
+    if (data.storage) {
+      version = parseInt(data.storage.version || 0, 10) + 1
+    }
+
+    while (migrations[version]) {
+      const newData = migrations[version](data)
+      localStorage.setItem(
+        key,
+        JSON.stringify({ ...(newData as Object), storage: { version } })
```
cazala: it looks like this line is getting rid of the `loading: true` property that comes with the initial state of the storage reducer (see this line https://github.com/decentraland/decentraland-dapps/pull/11/files#diff-8544c74e5fd9e4b18f35209ce5efd2beR12) I dunno if that breaks anything, probably @NicoSantangelo knows better than myself about this, but it seems like this should rather be:

```ts
JSON.stringify({ ...(newData as Object), storage: { ...newData.storage, version } })
```

cazala:
I left a question in the comments but it looks pretty good to me 👍

src/lib/localStorage.ts
NicoSantangelo: I think this is not a big deal because the `loading` will typically be set after creating the middleware, as a part of the redux-storage lib.
That said I think it doesn't hurt adding the spread, _just in case_

Also, can we get rid of the as `Object`? Maybe be a bit more specific?

README.md
```diff
@@ -305,15 +305,17 @@ You will need to create a `storageMiddleware` and add apply it along with your o
...
-const { storageMiddleware, loadStorageMiddleware } = createStorageMiddleware(
-  'storage-key', // this is the key used to save the state in localStorage (required)
-  [], // array of paths from state to be persisted (optional)
-  [] // array of actions types that will trigger a SAVE (optional)
-)
+const { storageMiddleware, loadStorageMiddleware } = createStorageMiddleware({
+  storageKey: 'storage-key' // this is the key used to save the state in localStorage (required)
+  paths: [] // array of paths from state to be persisted (optional)
+  actions: [] // array of actions types that will trigger a SAVE (optional)
+  migrations: migrations // migration object that will migrate your localstorage
```
cazala: this should be marked as `(optional)` too

README.md
```diff
@@ -290,12 +290,12 @@ function* handleFetchLandAmountRequest(action: FetchLandAmountRequestAction) {
 
 ## Storage
 
-The storage module allows you to save parts of the redux store in localStorage to make them persistent.
-This module is required to use other modules like `Transaction` and `Translation`.
+The storage module allows you to save parts of the redux store in localStorage to make them persistent and migrate it from different versions without loosing it.
+This module is required to use other modules like `Transaction`, `Translation`, `Wallet` and `Storage`.
```
cazala: The `Storage` module is irrelevant in this enumeration (this is already the `Storage` module)

README.md
```diff
@@ -325,6 +327,30 @@ const store = createStore(rootReducer, enhancer)
...
 
+**Migrations**:
+
+`migrations` looks like
+
+`migrations.ts`:
+
+```ts
+export const migrations = {
+  1: migrateToVersion1(data),
```
cazala: Does the `1` migration ever exist? doesn't it start from `2`?

README.md
```diff
@@ -347,7 +373,7 @@ export const rootReducer = storageReducerWrapper(
 
 ### Advanced Usage
 
-This module is necessary to use other modules like `Transaction` or `Translation`, but you can also use it to make other parts of your dApp's state persistent
+This module is necessary to use other modules like `Transaction`, `Translation`, `Wallet` and `Storage`, but you can also use it to make other parts of your dApp's state persistent
```
cazala: Same as the first comment ☝️

PR #12, opened 2018-09-13
feat: add method buildTransactionWithReceiptPayload

nachomazzara:
- Add `buildTransactionWithReceiptPayload`
- Add circle

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,13 +10,13 @@ jobs:
           command: sudo apt-get update && sudo apt install -y libusb-1.0-0 libusb-1.0-0-dev
       - restore_cache:
           keys:
-            - v1-dependencies-{{ checksum "package.json" }}
+            - v1-dependencies-{{ checksum "package-lock.json" }}
             - v1-dependencies-
       - run: npm i
       - save_cache:
           paths:
             - node_modules
-          key: v1-dependencies-{{ checksum "package.json" }}
+          key: v1-dependencies-{{ checksum "package-lock.json" }}
       - run: npm run test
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,26 @@
...
+    docker:
+      - image: circleci/node:9.11.1
+    steps:
+      - checkout
+      - run:
+          name: Install system dependencies
+          command: sudo apt-get update && sudo apt install -y libusb-1.0-0 libusb-1.0-0-dev
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "package.json" }}
```
NicoSantangelo: I think it's better to use `package-lock.json` to avoid cache conflicts

src/modules/transaction/types.ts
```diff
@@ -27,3 +30,7 @@ export enum NetworkName {
   kovan = 'kovan',
   rinkeby = 'rinkeby'
 }
+
+interface Receipt {
+  logs: any[]
```
NicoSantangelo: Is it possible to type this? or is it way too dynamic?

nachomazzara: It has events logs on it, `args` of events could be dynamic. 

I can do something like
```ts
logs: {
name: string,
args: any,
[key: string]: any
}
```

Is it worth?

NicoSantangelo: think so! it allows VSCode to autocomplete .name and .args at least

NicoSantangelo:
🎉

PR #19, opened 2018-09-25
chore: remove unsued dependencies

cazala:
This PR removes the system dependencies on `libusb` since they are not required anymore by `decentraland-eth`. It also allow us to move to node 10.


NicoSantangelo:
<3

PR #29, opened 2018-12-14
feat: location

cazala:
This PR adds common functionality for the `location` module regarding the sign in page and the home page.

It also adds an `<App>` container that can be used to easily bootstrap an application that basically combines the `Navbar`, `Footer` and `Page` components.

I also updated the README with this new stuff and added missing documentation from previous releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,10 @@ Common modules for our dApps
 - [Lib](https://github.com/decentraland/decentraland-dapps#lib)
   - [API](https://github.com/decentraland/decentraland-dapps#api)
 - [Containers](https://github.com/decentraland/decentraland-dapps#lib)
+  - [App](https://github.com/decentraland/decentraland-dapps#app)
   - [Navbar](https://github.com/decentraland/decentraland-dapps#navbar)
+  - [Footer](https://github.com/decentraland/decentraland-dapps#footer)
+  - [SignInPage](https://github.com/decentraland/decentraland-dapps#signinpage)
   - [EtherscanLink](https://github.com/decentraland/decentraland-dapps#etherscanlink)
 
 # Modules
@@ -929,17 +932,33 @@ Also, all the pending actions are stored in an array in `state.invite.loading` s
 
 ## Location
 
-The location module provides a `navigateTo` action that wrapps `react-router-redux`'s `push` action. It also provides some helpful selectors:
+The location module provides a `navigateTo`, `navigateToSignIn` and `navigateToRoot` actions that wrapps `react-router-redux`'s `push` action. It also provides some helpful selectors:
 
 ```ts
 getLocation(state)
 getPathname(state)
 getPathAction(state) // returns the final part of a url (after the last slash)
+isSignIn(state)
+isRoot(state)
 ```
 
 ### Installation
 
-You need to add a saga to use this module
+You need to add a reducer and a saga to use this module
+
+**Reducer**:
+
+Add the `translationReducer` as `translation` to your `rootReducer`:
+
+```ts
+import { combineReducers } from 'redux'
+import { locationReducer as location } from 'decentraland-dapps/dist/modules/location/reducer'
+
+export const rootReducer = combineReducers({
+  location
+  // your other reducers
+})
+```
 
 **Saga**:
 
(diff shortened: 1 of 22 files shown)


README.md
```diff
@@ -1026,6 +1072,275 @@ export default class Page extends React.PureComponent {
...
+
+This `<Footer>` will show all only English and Spanish as the options in the language dropdown. If you don't provide any it will use the 6 supported languages: `en`, `es`, `fr`, `jp`, `ko` and `zh`.
+
+### i18n
+
+If you are using the [Translation](https://github.com/decentraland/decentraland-dapps#translation) module, the `Footer` contatiner comes with support for the 6 languages supported by the library.
+
+### Advanced Usage
+
+You can override any of the default translations for any locale if you need to
```
nachomazzara: This is similar to the one we have in `Navbar` and other sections. What do you think of moving it to the translations section where you point the keys used?

README.md
```diff
@@ -955,6 +974,33 @@ export function* rootSaga() {
...
+import { combineReducers } from 'redux'
+import { createLocationReducer } from 'decentraland-dapps/dist/modules/location/reducer'
+
+const locationReducer = createLocationReducer({
+  root: '/',
+  signIn: '/sign-in'
+})
+
+export const rootReducer = combineReducers({
+  location
```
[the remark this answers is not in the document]
cazala: actually it has to be named `location` so the selectors work, so the line that was wrong is the one above where I instantiated it.

good catch!

README.md
cazala: I know it looks repetitive but I think as the library grows it is better to keep the translation options on its respective component rather than having them all in one place, otherwise at some point it might be difficult to tell apart which translation keys belong to each component

README.md
```diff
@@ -929,17 +932,33 @@ Also, all the pending actions are stored in an array in `state.invite.loading` s
 
 ## Location
 
-The location module provides a `navigateTo` action that wrapps `react-router-redux`'s `push` action. It also provides some helpful selectors:
+The location module provides a `navigateTo`, `navigateToSignIn` and `navigateToRoot` actions that wrapps `react-router-redux`'s `push` action. It also provides some helpful selectors:
```
NicoSantangelo:
```suggestion
The location module provides the `navigateTo`, `navigateToSignIn` and `navigateToRoot` actions that wrapp `react-router-redux`'s `push` action. It also provides some helpful selectors:
```

README.md
```diff
@@ -1026,6 +1072,275 @@ export default class Page extends React.PureComponent {
 
 This `<Navbar>` will show the user's blockie and mana balance because it is connected to the store.
 
+## Navbar
+
+The `<Navbar>` container can be used in the same way as the `<Navbar>` component from `decentaland-ui` but it's already connected to the redux store. You can override any `NavbarProp` if you want to connect differently, and you can pass all the regular `NavbarProps` to it.
```
NicoSantangelo: Maybe we should add a link to `[decentraland-ui](https://github.com/decentraland/ui)` here and in the other instances?

README.md
```diff
@@ -1026,6 +1072,275 @@ export default class Page extends React.PureComponent {
...
+          <Container>{children}</Container>
+        </div>
+        <Footer locales={['en', 'es']} />
+      </>
+    )
+  }
+}
+```
+
+This `<Footer>` will show all only English and Spanish as the options in the language dropdown. If you don't provide any it will use the 6 supported languages: `en`, `es`, `fr`, `jp`, `ko` and `zh`.
```
NicoSantangelo: I think `This <Footer> will show all only English and Spanish` sounds kind of weird. I'm not sure what it means

src/containers/App/App.container.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const mergeProps = (
+  stateProps: MapStateProps,
+  dispatchProps: object,
+  ownProps: AppProps
+) => ({
+  ...stateProps,
+  ...dispatchProps,
+  ...ownProps
+})
```
NicoSantangelo: Question: I don't get why is this here, isn't this equal to the default mergeProps?

src/modules/wallet/types.ts
```diff
@@ -24,6 +24,6 @@ export interface EthereumWindow {
     _metamask: { isApproved: () => Promise<boolean> }
     isApproved: () => Promise<boolean>
 
-    enable: () => Promise<string[]>
+    enable?: () => Promise<string[]>
```
NicoSantangelo: this can be undefined?

README.md
cazala: typo 😅 it should be "This `<Footer>` will show only English and Spanish..."

src/containers/App/App.container.tsx
cazala: nope, the default `mergeProps` is `Object.assign({}, ownProps, stateProps, dispatchProps)`, so you can't override the connected props

src/modules/wallet/types.ts
cazala: i dunno, but the linter was complaining about this line: https://github.com/decentraland/decentraland-dapps/blob/master/src/modules/wallet/utils.ts#L56

saying that the `typeof ethereum.enable === 'function'` was unnecessary because it was typed as a non-optional function

If we remove that part we can switch this back to be a required prop

README.md
cazala: makes sense, I'll add those 👍

PR #31, opened 2019-01-29
feat: add a modal provider

nicosantangelo:
TODO:
- [x] Fix dependency error with semantics Modal
- [x] Update README

Related: https://github.com/decentraland/ui/pull/26

### DIFF
--- a/README.md
+++ b/README.md
@@ -1009,7 +1009,7 @@ Leverages redux state and provides actions to open and close each modal by name.
 ```ts
 openModal(name: string, metadata: any = null)
 closeModal(name: string)
-closeAllModal()
+closeAllModals()
 ```
 
 It also provides a selector to get the open modals:
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ Common modules for our dApps
   - [Analytics](https://github.com/decentraland/decentraland-dapps#analytics)
   - [Loading](https://github.com/decentraland/decentraland-dapps#loading)
   - [Location](https://github.com/decentraland/decentraland-dapps#location)
+  - [Modal](https://github.com/decentraland/decentraland-dapps#modal)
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -1001,6 +1002,95 @@ This way you can change the default locations to use different ones. This will b
...
 </details>
 
+## Modal
+
+Leverages redux state and provides actions to open and close each modal by name. It provides two simple actions:
+
+```ts
+openModal(name: string, metadata: any = null)
+closeModal(name: string)
+closeAllModal()
```
cazala: closeAllModal`s`

src/containers/App/App.tsx
```diff
@@ -13,13 +13,13 @@ export default class App extends React.PureComponent<
     const { hero, isHomePage, children, ...rest } = this.props
     const hasHero = hero != null && isHomePage
     return (
-      <>
+      <React.Fragment>
```
cazala: Why this change?

README.md
nicosantangelo: Darn it!

src/containers/App/App.tsx
nicosantangelo: The only other piece of code I found using a Fragment used `React.Fragment` (TranslationProvider.tsx) so I went with that on ModalProvider and wanted to use the same thing everywhere

BUT

looking closer the README uses `<>` 😅 soo...which one do you like best? I'll update the PR only using one of the two. I have no preference

cazala: I would rather use `<>`, i think the marketplace is the only one that still uses `React.Fragment` but everywhere else i think we are using the short version (i like it better 😇 ).

nicosantangelo: <> it is 🚂 !

PR #35, opened 2019-02-19
feat: modal changes

nicosantangelo:
Modal features:
- Adhere more closely to the react lifecycle for modals. `componentWillUnmount` will fire when the modal closes
- Pass down a onClose prop to provider modals

Not sure the added complexity is worth it. We can keep only this commit: https://github.com/decentraland/decentraland-dapps/commit/7af6055206f531f3eeffd810539ed15c289c7e79 (which adds `componentWillUnmount`) if we want to remove it

### DIFF
--- a/src/containers/Modal/Modal.tsx
+++ b/src/containers/Modal/Modal.tsx
@@ -1,0 +1,25 @@
+import * as React from 'react'
+import { Modal as ModalComponent } from 'decentraland-ui'
+
+import { ModalProps } from './Modal.types'
+
+export default class Modal extends React.PureComponent<ModalProps> {
+  handleOnClose = () => {
+    const { name, onCloseModal } = this.props
+    onCloseModal(name)
+  }
+
+  render() {
+    const { name, ...modalProps } = this.props
+
+    return (
+      <ModalComponent
+        open={true}
+        className={name}
+        size="small"
+        onClose={this.handleOnClose}
+        {...modalProps}
+      />
+    )
+  }
+}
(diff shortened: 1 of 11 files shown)


src/containers/Modal/Modal.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react'
+import { Modal as ModalComponent } from 'decentraland-ui'
+
+import { ModalProps } from './Modal.types'
+
+export default class Modal extends React.PureComponent<ModalProps> {
+  handleOnClose = () => {
```
cazala: Following the [naming conventions](https://github.com/decentraland/standards/blob/master/standards/react-redux.md#21-components) this should be only `handleClose` 🤓

src/providers/ModalProvider/ModalProvider.tsx
```diff
@@ -1,29 +1,42 @@
...
-import { DefaultProps, Props } from './ModalProvider.types'
+import { DefaultProps, Props, ModalComponent } from './ModalProvider.types'
 
 export default class ModalProvider extends React.PureComponent<Props> {
   static defaultProps: DefaultProps = {
     children: null
   }
 
+  getOnClose(name: string) {
+    return () => this.props.onClose(name)
```
cazala: I think something like `getHandler` or `getCloseHandler` sounds more natural 🤓 🤓 🤓

nicosantangelo:
Super thanks for the review! applied the suggested changes

PR #43, opened 2019-06-17
fix: move some devDependencies to depencies to avoid version incompatibility

nachomazzara:
E.g: `redux-sagas@v1` has `delay` as a built-in effect and it was removed from the root.

https://redux-saga.js.org/docs/recipes/

Usage: https://github.com/decentraland/decentraland-dapps/blob/master/src/modules/transaction/sagas.ts#L10


nachomazzara:
Followed by #44

PR #49, opened 2019-07-25
fix: storage transform type

belohlavek:
It needs to use T defined in `StorageMiddleware<T>`


cazala:
Depends on https://github.com/decentraland/redux-persistence/pull/1 to be released and bumped here before merging

PR #68, opened 2020-03-18
feat: handle network and account changes

cazala:
This PR adds new functionality:

- New actions added: `CHANGE_ACCOUNT`, `CHANGE_NETWORK` and `DISCONNET_WALLET`

- `WalletProvider` dispatches `CHANGE_ACCOUNT` and `CHANGE_NETWORK` actions when a change is detected, so dApps can react to it.

- `disconnectWallet()` can be dispatched to go back to non-connected state.

Changes:

There are no breaking changes, although now it's not possible to set the `MANA_ADDRESS` from the outside. This renders `createWalletSaga` obsolete, but I left it with a deprecation warning encouraging users to import `walletSaga` directly instead.

This can be tested using the release candidate `decentraland-dapps@7.4.0-rc8`.

I deployed a version of the marketplace using this for easier manual testing here: https://public-930mdxzai.now.sh/

### DIFF
--- a/src/modules/wallet/actions.ts
+++ b/src/modules/wallet/actions.ts
@@ -27,3 +27,17 @@ export const enableWalletFailure = (error: string) =>
 export type EnableWalletRequestAction = ReturnType<typeof enableWalletRequest>
 export type EnableWalletSuccessAction = ReturnType<typeof enableWalletSuccess>
 export type EnableWalletFailureAction = ReturnType<typeof enableWalletFailure>
+
+export const CHANGE_ACCOUNT = 'Change Account'
+export const changeAccount = (address: string) =>
+  action(CHANGE_ACCOUNT, { address })
+export type ChangeAccountAction = ReturnType<typeof changeAccount>
+
+export const CHANGE_NETWORK = 'Change Network'
+export const changeNetwork = (network: number) =>
+  action(CHANGE_NETWORK, { network })
+export type ChangeNetworkAction = ReturnType<typeof changeNetwork>
+
+export const DISCONNECT_WALLET = 'Disconnect'
+export const disconnectWallet = () => action(DISCONNECT_WALLET)
+export type DisconnectWalletAction = ReturnType<typeof disconnectWallet>
--- a/src/modules/wallet/actions.ts
+++ b/src/modules/wallet/actions.ts
@@ -29,13 +29,13 @@ export type EnableWalletSuccessAction = ReturnType<typeof enableWalletSuccess>
 export type EnableWalletFailureAction = ReturnType<typeof enableWalletFailure>
 
 export const CHANGE_ACCOUNT = 'Change Account'
-export const changeAccount = (address: string) =>
-  action(CHANGE_ACCOUNT, { address })
+export const changeAccount = (wallet: Wallet) =>
+  action(CHANGE_ACCOUNT, { wallet })
 export type ChangeAccountAction = ReturnType<typeof changeAccount>
 
 export const CHANGE_NETWORK = 'Change Network'
-export const changeNetwork = (network: number) =>
-  action(CHANGE_NETWORK, { network })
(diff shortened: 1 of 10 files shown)


src/modules/wallet/actions.ts
```diff
@@ -27,3 +27,17 @@ export const enableWalletFailure = (error: string) =>
...
+  action(CHANGE_ACCOUNT, { address })
+export type ChangeAccountAction = ReturnType<typeof changeAccount>
+
+export const CHANGE_NETWORK = 'Change Network'
+export const changeNetwork = (network: number) =>
+  action(CHANGE_NETWORK, { network })
+export type ChangeNetworkAction = ReturnType<typeof changeNetwork>
+
+export const DISCONNECT_WALLET = 'Disconnect'
+export const disconnectWallet = () => action(DISCONNECT_WALLET)
```
nachomazzara: Do we have a real use case for this?

src/providers/WalletProvider/WalletProvider.ts
```diff
@@ -1,14 +1,73 @@
...
 
-  async UNSAFE_componentWillMount() {
+  UNSAFE_componentWillMount() {
+    // try to connect wallet
     const { onConnect } = this.props
     onConnect()
+
+    // add listeners
+    this.on('accountsChanged', this.handleChangeAccount)
+    this.on('networkChanged', this.handleChangeNetwork)
```
nachomazzara: Based on this [EIP](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#chainchanged), maybe we should add `chainChanged` too. Provider builders are not following any finished standard yet.

Also, these methods should send a parameter when executing the listener related to what happened

```typescript
ethereum.on('chainChanged', listener: (chainId: Integer) => void): this;
```

and

```typescript
ethereum.on('accountsChanged', listener: (accounts: Array<String>) => void): this;
```

In case you want to check first them before calling the provider again.

src/providers/WalletProvider/WalletProvider.ts
```diff
@@ -1,14 +1,73 @@
...
+
+  handle(
+    action: 'on' | 'removeListener',
+    type: 'accountsChanged' | 'networkChanged',
+    handler: Function
+  ) {
+    // try to use web3x abstraction
+    if (this.eth) {
+      try {
+        this.eth.provider[action](type as any, handler as any)
```
nachomazzara: It is showing an error if we type `type` and `handler`, right?

src/modules/wallet/actions.ts
cazala: I dunno if it's that important, but the current builder designs that uses Auth0 has a Sign Out button. The idea is to get rid of Auth0 in favor of ephemeral keys, but the designs would be the same, so I guess the Sign Out button would use this.

src/providers/WalletProvider/WalletProvider.ts
cazala: yes, because of the dynamic accessor `action` it can't be properly typed.

src/providers/WalletProvider/WalletProvider.ts
cazala: `chainChanged` is not supported by web3x, it uses `networkChanged`. Also MetaMask advices to use `networkChanged` on its warning message.

```
MetaMask will soon stop reloading pages on network change.
If you rely upon this behavior, add a 'networkChanged' event handler to trigger the reload manually: https://metamask.github.io/metamask-docs/API_Reference/Ethereum_Provider#ethereum.on(eventname%2C-callback)
Set 'ethereum.autoRefreshOnNetworkChange' to 'false' to silence this warning: https://metamask.github.io/metamask-docs/API_Reference/Ethereum_Provider#ethereum.autorefreshonnetworkchange'
```

cazala:
@nachomazzara thanks for the feedback! I addressed it in my last commit.

PR #77, opened 2020-07-20
fix: add samsung copies

2fd:


src/modules/translation/defaults/ja.json
```diff
@@ -41,7 +41,8 @@
       "get_started": "開始する",
       "options": {
         "desktop": "{metamask_link}拡張や{ledger_nano_link}のようなハードウェアウォレットを使用できます。",
-        "mobile": "{coinbase_link}や{imtoken_link}などのモバイルブラウザを使用できます。"
+        "mobile": "{coinbase_link}や{imtoken_link}などのモバイルブラウザを使用できます。",
+        "samsung": "You can use {samsung_link}."
```
nachomazzara, line 45: I would use google translator or ask Waann Chin for translations

PR #82, opened 2020-09-01
fix: don't mark tx as REVERTED when in block, check tx receipt status instead.

### LINKED ISSUE #81: Wrongly marking a transaction as "reverted" when it didn't
Running some tests on Ropsten network, seems to happen when the network is slow, a transaction is being wrongly marked as Reverted status, when it didn't.

For example, tx hash [0x3cc266a2d188e26eed3033f34eafa8f7f583d40f09db76dfb8430557ddb23a66](https://ropsten.etherscan.io/tx/0x577477ce78b5309bc3fa37f7e5582515bebb5ed0468979dfdac8936f9a73d9e1), took about 1-2 mins to get up to 4 block confirmations. 

When debugging this, I've noticed that the response of `await eth.getTransaction(hash)`, was including a defined `blockNumber` value, which means it has been "received by the network" - ie. [not pending/queued/replaced](https://github.com/decentraland/decentraland-dapps/blob/776b86012183890b5ccd7ddb23dcc01b92ffe5fb/src/modules/transaction/txUtils.ts#L53). **But** later on, the call to `await eth.getTransactionReceipt(hash)` still returned `null` because the tx was still _pending_, ie. the block was not yet mined, according to web3 docs https://web3js.readthedocs.io/en/v1.2.0/web3-eth.html#gettransactionreceipt .

However in the [next conditional](https://github.com/decentraland/decentraland-dapps/blob/776b86012183890b5ccd7ddb23dcc01b92ffe5fb/src/modules/transaction/txUtils.ts#L86:L93), the if `receipt == null` considers it as being "reverted" which is not correct.

tmilar:
Fixes #81.

### DIFF
--- a/src/modules/transaction/txUtils.ts
+++ b/src/modules/transaction/txUtils.ts
@@ -83,11 +83,21 @@ export async function getTransaction(
 
   const receipt = await eth.getTransactionReceipt(hash)
 
+  if (receipt == null) {
+    // pending (in block, but not yet mined)
+    const tx: PendingTransaction = {
+      status: TransactionStatus.PENDING,
+      ...response
+    }
+    return tx
+  }
+
   // reverted
-  if (receipt == null || !receipt.status) {
+  if (!receipt.status) {
     const tx: RevertedTransaction = {
       status: TransactionStatus.REVERTED,
-      ...response
+      ...response,
+      receipt
     }
     return tx
   }
--- a/src/modules/transaction/types.ts
+++ b/src/modules/transaction/types.ts
@@ -33,6 +33,7 @@ export type PendingTransaction = TransactionResponse & {
 
 export type RevertedTransaction = TransactionResponse & {
   status: TransactionStatus.REVERTED
+  receipt: TransactionReceipt
 }
 
 export type ConfirmedTransaction = TransactionResponse & {


src/modules/transaction/txUtils.ts
```diff
@@ -83,11 +83,21 @@ export async function getTransaction(
 
   const receipt = await eth.getTransactionReceipt(hash)
 
+  if (receipt == null) {
```
nachomazzara, line 86: I'm not sure about this, sometimes the tx can be dropped or reorg leaving it outside the main-chain. If I'm not wrong, the receipt will `null` too. Leaving a forever loading. 

Maybe if `receipt == null` we can also check for something else which we are doing at line 53.

What was the case that motivates the PR?

tmilar: If it would get dropped, then wouldn't it be checked later on and moved from "pending" to "dropped" status?

Motivation for this PR: I was sending transactions in my app and listening to later actions FETCH_TRANSACTION_SUCCESS and FETCH_TRANSACTION_FAILURE, to update/refresh my app status accordingly. It became a problem though when I was 
 wrongly having FETCH_TRANSACTION_FAILURE as if they were reverted, when they actually wasn't.

nachomazzara: Cool! do you have examples? 

`Dropped & replaced` yes, will happen what you said. But it was reorg and not included in the chain the tx will be a ghost one like it will never happen.

tmilar: I was just creating a transaction, and listening to updates by using selectors to grab pending/confirmed transactions from the redux state.

Not sure how to handle the reorg case. In this scenario, what would be the expected result? Following the original code logic + your logic, I think you imply that it should be marked as Reverted?  However maybe that is a different issue, in order to be sure if the tx was actually placed in the main-chain (ie. "confirmed ~100%"?) , there should be some check of the # of confirmations, and act on the tx result after asserting that enough confirmations have been achieved. 

The current change on this PR, just keeps the tx status as "Pending" (instead of wrongly "Reverted"), until more information is available to say otherwise.  Should this change also include logic about the block hash and # confirmations? This [SO answer](https://ethereum.stackexchange.com/a/53270) proposes a strategy for achieving this

nachomazzara: > Should this change also include logic about the block hash and # confirmations? This SO answer proposes a strategy for achieving this.

This is exactly what I was trying to explain.

nachomazzara: @tmilar is this still in your backlog?

tmilar: Hi @nachomazzara , I'm sorry but I'm not actively working on this project anymore. 
Though if it still matters to you let me know, I can make an effort to complete the PR when I have some free time.

nachomazzara: It would be great if you can complete it :)

PR #93, opened 2020-12-29
feat: LoginModal

nicosantangelo:
# ⚠️ 
@cazala should this be a major release?
We're changing the way the SignInPage works (besides adding the new LoginModal Component), there's no longer a single way to connect your wallet


cazala (approved):
Awesome 🚀 the code is looking great, but let's add a `BREAKING CHANGE: <reason>` comment to at least 1 commit message so semantic-release can pick it up and make a major release, since we changed the way the `enableWalletRequest` action works.

src/modules/translation/defaults/en.json
```diff
@@ -44,6 +44,18 @@
         "mobile": "You can use mobile browsers such as {coinbase_link} or {imtoken_link}.",
```
nicosantangelo, line 44: You think we should update these options? the sign in page still reads 
```json
"desktop": "You can use the {metamask_link} extension or a hardware wallet like {ledger_nano_link}.",
```
@cazala

cazala: good catch, i think we could update it to something like `You can use the {metamask_link} extension, a hardware wallet like {ledger_nano_link} or login using your email account via {fortmatic_link}.` wdyt?

PR #98, opened 2021-01-28
feat: expose pristine provider

nachomazzara:
Expose the provider connected with `dcl-connect`. It is not trivial for the user to import  `connection` from `decentraland-connect`.


nicosantangelo (approved):
LGTM

PR #103, opened 2021-02-10
feat: mana L2

cazala:
Adds matic L2 to the wallet module and connects the new UserMenu

This is still using a release candidate version of `decentraland-ui`


src/containers/UserMenu/UserMenu.container.ts
```diff
@@ -0,0 +1,61 @@
...
+    hasActivity: getTransactions(state, address || '').some(tx =>
+      isPending(tx.status)
+    ),
+    hasTranslations: isEnabled(state)
+  }
+}
+
+const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
+  onSignIn: () => dispatch(disconnectWallet()),
+  onSignOut: () => dispatch(connectWalletRequest())
```
nicosantangelo, line 44: I these two are flipped!
Also, should the sign in action be an enable/login modal?

src/lib/graph.ts
```diff
@@ -0,0 +1,34 @@
...
+  query: string,
+  retryDelay = 500,
+  retryAttempts = 10
+) {
+  try {
+    const result: { data: T } = await fetch(url, {
+      method: 'post',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        query
```
nicosantangelo, line 12: I think you're stringifying a string!

src/containers/UserMenu/UserMenu.container.ts
cazala: Fixed! I also removed the mapping to the onSignIn, since actually many things could happen there, most of our dapps just take you to a `/sign-in` page for example. But it's not very important since we don't do the sign in via this component (we just use it once the user is already signed in). The actual sign in button that users click is the one thats part of the Navbar component

src/lib/graph.ts
cazala: mm i'm stringifying an object with a `query` string inside, or are you saying something else?

nicosantangelo (approved):

PR #118, opened 2021-03-08
feat: refresh wallet balances

cazala:
Also this PR removes the `balance` from `NetworkData` to avoid using Infura.

### DIFF
--- a/src/modules/wallet/reducer.ts
+++ b/src/modules/wallet/reducer.ts
@@ -59,40 +59,49 @@ export function walletReducer(
   action: WalletReducerAction
 ): WalletState {
   switch (action.type) {
-    case CONNECT_WALLET_REQUEST:
+    case FETCH_WALLET_REQUEST:
+    case ENABLE_WALLET_REQUEST:
+    case CONNECT_WALLET_REQUEST: {
       return {
         ...state,
         loading: loadingReducer(state.loading, action)
       }
-    case CONNECT_WALLET_SUCCESS:
+    }
+
+    case FETCH_WALLET_SUCCESS:
+    case CONNECT_WALLET_SUCCESS: {
       return {
         ...state,
         loading: loadingReducer(state.loading, action),
         error: null,
         data: action.payload.wallet
       }
-    case CONNECT_WALLET_FAILURE:
+    }
+
+    case FETCH_WALLET_FAILURE:
+    case CONNECT_WALLET_FAILURE: {
       return {
         ...state,
         loading: loadingReducer(state.loading, action)
       }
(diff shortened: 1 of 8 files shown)


src/modules/wallet/reducer.ts
```diff
@@ -98,6 +107,25 @@ export function walletReducer(
...
+      return {
+        ...state,
+        loading: loadingReducer(state.loading, action)
+      }
+    }
+    case FETCH_WALLET_SUCCESS: {
+      return {
+        ...state,
+        loading: loadingReducer(state.loading, action),
+        data: action.payload.wallet
```
nachomazzara, line 120: Should we need to clean `error`? If so, Can we move this with `ENABLE_WALLET_SUCCESS` ?

src/modules/wallet/reducer.ts
```diff
@@ -98,6 +107,25 @@ export function walletReducer(
         data: null
       }
     }
+    case FETCH_WALLET_REQUEST: {
```
nachomazzara, line 110: Is it possible to move all the `_REQUEST`  to the same case?

```
case CONNECT_WALLET_REQUEST
case ENABLE_WALLET_REQUEST
case FETCH_WALLET_REQUEST:
```

src/modules/wallet/reducer.ts
```diff
@@ -98,6 +107,25 @@ export function walletReducer(
...
+      return {
+        ...state,
+        loading: loadingReducer(state.loading, action),
+        data: action.payload.wallet
+      }
+    }
+    case FETCH_WALLET_FAILURE: {
+      return {
+        ...state,
+        loading: loadingReducer(state.loading, action)
```
nachomazzara, line 126: Should we need to return the `error` here? If so, Can we move this with `ENABLE_WALLET_FAILURE`?

nachomazzara (changes requested):
Looks good

src/modules/wallet/reducer.ts
cazala: the `error` is not populated nor cleaned by `CONNECT_WALLET_FAILURE` either, the reason is otherwise you would see an error as soon as you arrive to the login page. The only one that produces an error is `ENABLE_WALLET_FAILURE` because that one is not expected to failed, on the other hand `CONNECT_WALLET_FAILURE` is expected to happen when the app boots and you don't have a wallet connected, which just means you need to sign in.

src/modules/wallet/reducer.ts
cazala: Nope, explained above.

nachomazzara (approved):

PR #129, opened 2021-04-09
Feat: Modal when the user is connected in a wrong network.

### LINKED ISSUE #127: Display a message when the user is in a wrong network
If the user is at .io the network should be `ropsten`, `goerli` or `mumbai`
If the user is at `.stg` or `.org` the network should be `mainnet` or `matic`

malaniz:
Feat: Modal when the user is connected in a wrong network.
img-7VGumr_000

Closes #127

### DIFF
--- a/src/containers/Navbar/Navbar.types.tsx
+++ b/src/containers/Navbar/Navbar.types.tsx
@@ -1,6 +1,13 @@
 import { ChainId } from '@dcl/schemas'
 import { NavbarProps as NavbarComponentProps } from 'decentraland-ui'
 
+export declare type WrongNetworkModalI18N = {
+  wrongNetwork: {
+    header: React.ReactNode
+    message: React.ReactNode
+  }
+}
+
 export type NavbarProps = NavbarComponentProps & {
   chainId: ChainId | undefined
   hasTranslations?: boolean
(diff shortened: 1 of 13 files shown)


src/modules/translation/defaults/en.json
```diff
@@ -31,6 +31,10 @@
         "builder": "Builder",
         "events": "Events",
         "dao": "DAO"
+      },
+      "wrongNetwork": {
+        "header": "Wrong Network",
+        "message": "You need to be connected to {expectedChainName} Network to use this app, but you are currently connected to {currentChainName} Network."
```
nicosantangelo, line 37:
```suggestion
        "message": "You need to be connected to the {expectedChainName} Network to use this app, but you are currently connected to {currentChainName}."
```

src/containers/Navbar/Navbar.types.tsx
```diff
@@ -1,12 +1,26 @@
+import { ChainId } from '@dcl/schemas'
 import { NavbarProps as NavbarComponentProps } from 'decentraland-ui'
 
+export declare type WrongNetworkModalI18N = {
```
nicosantangelo, line 4:
```suggestion
export type WrongNetworkModalI18N = {
```

nicosantangelo (changes requested):
I understand that we're using the base components from decentraland-UI but we still need to adjust margins and paddings a bit

image

nachomazzara (approved):

PR #139, opened 2021-05-12
feat: allow for contractNames to be explicitly supplied

nicosantangelo:


src/modules/authorization/sagas.ts
```diff
@@ -165,11 +167,26 @@ async function changeAuthorization(
...
+      try {
+        contract = getContract(
+          getContractName(tokenAddress.toString()),
+          chainId
+        )
+      } catch (error) {
+        // We're defaulting to the ERC721CollectionV2 contract here as a last resort only if the wallet is trying to set approval for all of an ERC721 (which a collection wearable is)
+        // This is handling all of our use cases at the moment, and has virtually no downside because the transaction will fail regardless if the token contract address wasn't a collection.
+        // A more robust implementation is needed if we are going to support different kinds of ERC721 approvals
+        if (authorization.type === AuthorizationType.APPROVAL) {
```
nachomazzara, line 180: I'm imaging a use case that is not too crazy to happen: 

- If for some reason we need to deploy a new collection implementation because of a feature/vulnerability we will also need to remember to change this.

How much it cost to do something that solves this too?

nachomazzara (changes requested):

cazala (approved):

PR #151, opened 2021-06-02
feat(Authorization): configure meta-tx server url

cazala:
Added a `createAuthorizationSaga` function that accepts an optional configuration object `{ metaTransactionServerUrl?: string }` that allows to configure the `serverUrl` parameter of the `sendMetaTransaction` method used internally by the `authorization` module. The `authorizationSaga` is still exported with a non-configured meta transaction server. I figured is not needed for dapps that only run on Ethereum, but in case we want to make it mandatory to configure it we can remove that exported saga or replace it with a function that throws (that would require a major release instead of a minor one).

### DIFF
--- a/README.md
+++ b/README.md
@@ -578,11 +578,16 @@ export const rootReducer = combineReducers({
 
 **Sagas**
 
-Add the `authorizationSaga` to the `rootSaga`:
+Use `createAuthorizationSaga` and add the returned saga to the `rootSaga`:
 
 ```ts
 import { all } from 'redux-saga/effects'
-import { authorizationSaga } from 'decentraland-dapps/dist/modules/authorization/sagas'
+import { createAuthorizationSaga } from 'decentraland-dapps/dist/modules/authorization/sagas'
+
+// if you are not going to use meta transactions you can skip the options or just import `authorizationSaga`
+const authorizationSaga = createAuthorizationSaga({
+  metaTransactionServerUrl: 'https://transactions-api.decentraland.io'
+})
 
 export function* rootSaga() {
   yield all([
--- a/README.md
+++ b/README.md
@@ -578,7 +578,7 @@ export const rootReducer = combineReducers({
 
 **Sagas**
 
-Use `createAuthorizationSaga` and add the returned saga to the `rootSaga`:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -578,11 +578,16 @@ export const rootReducer = combineReducers({
...
 
-Add the `authorizationSaga` to the `rootSaga`:
+Use `createAuthorizationSaga` and add the returned saga to the `rootSaga`:
 
 ```ts
 import { all } from 'redux-saga/effects'
-import { authorizationSaga } from 'decentraland-dapps/dist/modules/authorization/sagas'
+import { createAuthorizationSaga } from 'decentraland-dapps/dist/modules/authorization/sagas'
+
+// if you are not going to use meta transactions you can skip the options or just import `authorizationSaga`
```
nicosantangelo, line 587: Maybe we can put this above as part of the explanation? Maybe it's easy to miss here (not that anyone reads READMEs anyways :P)

src/modules/authorization/sagas.ts
```diff
@@ -30,147 +30,163 @@ import {
...
 import { Authorization, AuthorizationAction, AuthorizationType } from './types'
 
-export function* authorizationSaga() {
-  yield takeEvery(
-    FETCH_AUTHORIZATIONS_REQUEST,
-    handleFetchAuthorizationsRequest
-  )
-  yield takeEvery(GRANT_TOKEN_REQUEST, handleGrantTokenRequest)
-  yield takeEvery(REVOKE_TOKEN_REQUEST, handleRevokeTokenRequest)
+export type AuthorizationSagaOptions = {
```
nicosantangelo, line 33: To be consistent, what do you think about moving this to /types as we have on wallet?

nicosantangelo (commented):
Comment

src/modules/authorization/sagas.ts
cazala: aah i didn't notice that on the wallet module, I made it like this by looking at the `translation` module 🦆 I'll move both to `/types` to keep it consistent.

nicosantangelo: 🏌️‍♂️

README.md
cazala: done ✅

src/modules/authorization/sagas.ts
cazala: done ✅

nicosantangelo (approved):

PR #153, opened 2021-06-03
feat: use multicall for authorizations

cazala:
I had to refactor the saga to use `ethers` instead of `web3x` because it didn't get along with the multicall provider. I didn't have to install `ethers` tho because it was already a dependency of other dependencies in the project.

I had to install also `@0xsequence/relayer` because otherwise the build would fail, although it's not explicitly imported in the code (it's an undocumented peer dependency of `@0xsequence/multicall`

I tested this as a release candidate `decentraland-dapps@12.10.0-rc7` installed in the marketplace using the production `.env` vars.

### DIFF
--- a/src/modules/authorization/sagas.ts
+++ b/src/modules/authorization/sagas.ts
@@ -121,7 +121,7 @@ export function createAuthorizationSaga(options?: AuthorizationSagaOptions) {
 
       const authorizationsToStore: Authorization[] = yield call(async () => {
         const results = await Promise.all(promises)
-        return results.filter(result => !!result) // filter nulls, or undefineds due to caughted promises
+        return results.filter(result => !!result) // filter nulls, or undefineds due to caught promises
       })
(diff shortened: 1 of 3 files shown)


src/modules/authorization/sagas.ts
```diff
@@ -50,48 +52,78 @@ export function createAuthorizationSaga(options?: AuthorizationSagaOptions) {
...
         const { chainId } = authorization
-        const address = Address.fromString(authorization.address)
-        const contractAddress = Address.fromString(
-          authorization.contractAddress
-        )
-        const authorizedAddress = Address.fromString(
-          authorization.authorizedAddress
-        )
 
+        // provider party 🎉
```
nachomazzara, line 63: 😆

nachomazzara (approved):

nicosantangelo (approved):

PR #159, opened 2021-07-20
feat: Add all the methods needed to manipulate the profile entity

LautaroPetaccio:
This PR does the following:
- Adds a lib (or a class) with the methods needed to create, authorize and deploy an entity based on the requirement of deploying an entity.
- Adds the `setProfileAvatarDescription[Request|Success|Failure]` and the `clearProfileError` actions alongside a handler in the sagas to perform the entity update (or deployment).
- Modifies the profile reducer to handle the new actions and to record the errors in case of a failure.
- Edits the README to provide documentation about how to use the modules and libraries added.


src/lib/entities.ts
```diff
@@ -0,0 +1,123 @@
...
+   * deploy an entity.
+   *
+   * @param address - The address of the deployer of the entity.
+   * @param entityId - The entity id that it's going to be deployed.
+   */
+  private async authenticateEntityDeployment(
+    address: string,
+    entityId: string
+  ): Promise<AuthChain> {
+    const eth: Eth = await getEthereumProvider()
```
cazala, line 61: We already have a `getNetworkProvider(chainId: ChainId)` to get a provider from any network, or `getConnectedProvider()` to get the provider that the user has connected (disregarding on which network it is connected to).

I would advise against adding a new helper to get a provider, I know this helper returns already a `web3x` eth abstraction, but since we plan to move away from web3x sooner than later, I'd use the `getConnectedProvider` helper and instantiate `Eth` here instead for the signature, like:

```ts
const provider = await getConnectedProvider()
const eth = new Eth(provider!)
```

src/lib/peer.ts
```diff
@@ -1,24 +1,38 @@
...
     }
-    const promise = this.lambdasClient
-      .fetchProfiles([address.toLowerCase()])
-      .then(profiles => profiles[0])
-    this.cache[address] = promise
-    return promise
+    const profiles = await this.lambdasClient.fetchProfiles([
+      address.toLowerCase()
+    ])
+    this.cache[address] = profiles[0]
```
cazala, line 23: This is caching _after_ the response has come back from the catalyst. The way it was done before (storing the promise in the cache) was so if two requests for the same profile happen simultaneously, only one request would go to the server.

This is very common when using the `<Profile />` container ([link](https://github.com/decentraland/decentraland-dapps/tree/master/src/containers/Profile)), pages like this one would send several requests for the same profile:

Screen Shot 2021-07-27 at 11 57 46 AM

src/modules/profile/sagas.ts
```diff
@@ -43,5 +57,62 @@ export function createProfileSaga({ peerUrl }: CreateProfileSagaOptions) {
...
+   * @param action - The action that triggered the handler.
+   */
+  function* handleSetProfileDescription(
+    action: SetProfileAvatarDescriptionRequestAction
+  ) {
+    try {
+      const { address, description } = action.payload
+
+      const entity: ProfileEntity = yield call(
+        [entities, 'getProfileEntity'],
```
cazala, line 75: Just out of curiosity, why using this notation instead of just `entities.getProfileEntity`? It seems like a more complicated way of doing the same, i dunno what am i missing.

LautaroPetaccio: This is [the proper way](https://redux-saga.js.org/docs/api/#callcontext-fn-args) of using the call method with a context. As we're calling the `getProfileEntity` method from the `entities` instance, we need to set the binding to the instance of that class for the call to work. I've noted that we used `yield call(() => x())` in some places, but using it that way is not how it's meant to be used, as it makes it hard to test because the call object being built has a reference to the lambda function instead of the method that we want to check for.

src/lib/entities.ts
LautaroPetaccio: Done!

src/lib/peer.ts
LautaroPetaccio: Ohhh, great catch! I didn't know that it was built that way to prevent multiple requests from happening.

src/modules/profile/sagas.ts
cazala: Gotcha, thanks for the explanation :)

src/lib/peer.ts
```diff
@@ -4,24 +4,30 @@ import { BaseAPI } from './api'
...
     }
-    const profiles = await this.lambdasClient.fetchProfiles([
-      address.toLowerCase()
-    ])
-    this.cache[address] = profiles[0]
-    return profiles[0]
+
+    this.cache[address] = this.lambdasClient
+      .fetchProfiles([address.toLowerCase()])
+      .then()
```
cazala, line 28: Is this last `.then()` necessary?

LautaroPetaccio: Oh yes! It is! Sorry, I forgot to write the whole line. We need it so we can retrieve only one profile

cazala (approved):

PR #162, opened 2021-08-05
fix: Avatar description update

LautaroPetaccio:
This PR changes the way that the profile is updated after a description change in an avatar.
The idea of this change is to re-use the information we already have without the need of doing another request to the profile lambda endpoint.


src/modules/profile/sagas.spec.ts
```diff
@@ -54,7 +54,7 @@ describe('when handling the action to set the profile avatar description', () =>
   })
 
   describe('when the deployment is successful', () => {
-    it('should deploy the new entity with the decription and the version changed', () => {
+    it.only('should deploy the new entity with the decription and the version changed', () => {
```
cazala, line 57: this `only` should be removed right?

cazala (approved):

PR #165, opened 2021-08-16
feat: Update React, DCL UI and others

LautaroPetaccio:
This PR updates DCL-UI, React and React-intl that depends on react.
There's also a small types fix that was breaking the marketplace.


nicosantangelo (approved):
LGTM

PR #166, opened 2021-08-18
feat: support provider on different chain

cazala:
This PR adds several helper to allow our dapps to support native transactions on networks that currently are only supported via meta-transactions.

## Utils
+ Helper to send transactions within sagas, that automatically decides to either send a regular or a meta transaction:

```ts
function* sendWalletTransaction(
  contractData: ContractData, 
  getTransaction: (contractInstance: ethers.Contract) => TransactionRequest
)
```

Example:

```ts
const contract = getContract(ContractName.MANAToken, ChainId.MATIC_MAINNET)
const hash = yield sendWalletTransaction(contract, mana => mana.transfer('0x...', toWei(1)))
```

+ Helpers to disable certain parts of the UI:

`CheckChain` checks if a `chainId` is supported by the current provider. It always renders the children, and it provides a boolean `isSupported` to control the state of the children. If the chain is not supported, it adds a tooltip to the children pointing the user to the right network:

```tsx
<CheckChain chainId={ChainId.MATIC_MUMBAI}>
{isSupported => (...)}
</CheckChain>
```

Example:
Screen Shot 2021-08-18 at 4 59 55 PM

`<ChainButton>` is a wrapper of decentraland-ui's `<Button>` with an extra prop `chainId` that uses internally the `CheckChain` component from above. Since buttons are a common use case, and disabling a button needs special treatment (the disabled styles currently don't allow for a popup to work on a disabled button), I figured it would be handy to have a dedicated component for this.

```tsx
<ChainButton chainId={ChainId.MATIC_MUMBAI} onClick={...}/>Do Something</ChainButton>
```

Screen Shot 2021-08-18 at 5 00 04 PM

+ Also this PR adds the ability to change networks via a `switchNetworkRequest(chainId: ChainId)` action, and it added the functionality to the exisitng `WrongNetworkModal` to switch to the right network:

Screen Shot 2021-08-18 at 5 00 46 PM

I also added a new "partial support" modal to tell the user that they won't be able to use all the features in the dapp, and that they will need to pay for transaction costs:

Screen Shot 2021-08-18 at 5 00 21 PM

You can test these features in this deployment of the account dapp: https://account-gvglekji0-decentraland1.vercel.app/


src/modules/wallet/utils.ts
```diff
@@ -60,7 +76,124 @@ export async function buildWallet(): Promise<Wallet> {
...
     chainId
   }
 }
+
+async function getTargetNetworkProvider(chainId: ChainId) {
+  const networkProvider = await getNetworkProvider(chainId)
+  return new providers.Web3Provider(networkProvider)
+}
+
+export function* sendWalletTransaction(
```
LautaroPetaccio, line 89: Would you like to test this method?

src/modules/wallet/utils.ts
```diff
@@ -1,15 +1,29 @@
...
 } from '../../lib/eth'
 import { getChainConfiguration } from '../../lib/chainConfiguration'
-import { Networks, Wallet } from './types'
+import { AddEthereumChainParameters, Networks, Wallet } from './types'
+import { getAddress, getChainId } from './selectors'
+
+let TRANSACTIONS_API_URL = 'https://transactions-api.decentraland.co/v1'
+export const getTransactionsApiUrl = () => TRANSACTIONS_API_URL
+export const setTransactionsApiUrl = (url: string) =>
+  (TRANSACTIONS_API_URL = url)
```
LautaroPetaccio, line 26: This was left hardcoded here, we might want to get it from the configs.

src/lib/eth.ts
```diff
@@ -56,3 +61,12 @@ export function isDapperProvider() {
 export function isValidChainId(chainId: string | number) {
   return Object.values(ChainId).includes(Number(chainId))
 }
+
+export function getChainIdByNetwork(network: Network) {
+  const connectedChainId = getConnectedProviderChainId()
+  if (!connectedChainId) {
+    throw new Error('Could not get connected provider chain id')
+  }
+  const config = getChainConfiguration(connectedChainId)
```
LautaroPetaccio, line 70: This method doesn't fail if the chain configuration isn't found for that chain id. What's the desired output of this function?

LautaroPetaccio (changes requested):

src/containers/ChainButton/ChainButton.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import { Button as ButtonComponent } from 'decentraland-ui'
+import { Props } from './ChainButton.types'
+import ChainCheck from '../ChainCheck'
+
+const disabledStyle = { opacity: 0.5, cursor: 'not-allowed', transform: 'none' }
```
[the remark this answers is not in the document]
cazala: We don't have CSS files on `decentraland-dapps`, and the current disabled styles of the `Button` don't let the `Popup` button work, because of the `pointer-events: none` rule.

src/modules/wallet/sagas.ts
```diff
@@ -157,6 +171,49 @@ function* handleConnectWalletSuccess() {
...
+  const provider: Provider | null = yield call(getConnectedProvider)
+  try {
+    if (!provider) {
+      throw new Error('Could not get provider')
+    }
+    const value: null = yield call([provider, 'request'], {
+      method: 'wallet_switchEthereumChain',
+      params: [{ chainId: '0x' + chainId.toString(16) }]
+    })
+    console.log('switch value', value)
```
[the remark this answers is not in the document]
cazala: Removed 👍

src/modules/wallet/sagas.ts
```diff
@@ -157,6 +171,49 @@ function* handleConnectWalletSuccess() {
...
+    yield put(switchNetworkSuccess(chainId))
+  } catch (switchError) {
+    // This error code indicates that the chain has not been added to MetaMask.
+    if (provider && switchError.code === 4902) {
+      try {
+        const value: null = yield call([provider, 'request'], {
+          method: 'wallet_addEthereumChain',
+          params: [getAddEthereumChainParameters(chainId)]
+        })
+        console.log('add value', value)
```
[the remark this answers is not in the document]
cazala: Removed 👍

src/lib/eth.ts
cazala: `getChainConfiguration(chainId: ChainId): ChainConfiguration` takes a `chainId` and always return a `ChainConfiguration` for all the values of the enum `ChainId`, so it can't return `null` or throw. 

This helper is used to get the `chainId` of a given `Network`. For instance, on the Account dapp we have several parts where we have the `Network` but we don't know the `chainId`, with this helper we can do:

```tsx
<ChainButton chainId={getChainIdByNetwork(network)}>
```

src/modules/translation/defaults/fr.json
```diff
@@ -32,9 +35,15 @@
...
-      "wrongNetwork": {
+      "wrong_network": {
         "header": "Mauvais réseau",
         "message": "Vous devez être connecté au réseau {expectedChainName} pour utiliser cette application, mais vous êtes actuellement connecté au réseau {currentChainName}."
+      },
+      "partially_supported_network": {
+        "header": "Réseau partiellement pris en charge",
+        "message": "Vous devez être connecté à {expectedChainName} pour utiliser toutes les fonctionnalités de cette application. De cette façon, toutes les transactions qui devraient se produire sur d'autres réseaux seraient envoyées en tant que méta-transactions. Étant donné que vous êtes actuellement connecté à {currentChainName}, vous ne pourrez utiliser que les fonctionnalités de cette application qui interagissent avec ce réseau particulier, et vous devrez payer les coûts de gaz de ces transactions.",
+        "switch_button": "Passer à {chainName}",
+        "continue_button": "Continuer sur {chainName}"
```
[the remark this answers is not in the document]
cazala: good catch! fixed it :)

src/containers/ChainButton/ChainButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export default class ChainButton extends React.PureComponent<Props> {
```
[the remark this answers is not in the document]
cazala: I stuck to the class since all the components in this repo are class components tbh, since that's the approach we are taking with the other repos (marketplace, builder, account).

src/containers/ChainCheck/ChainCheck.types.ts
```diff
@@ -0,0 +1,6 @@
+import { ChainId } from '@dcl/schemas'
+
+export type Props = {
+  chainId: ChainId
+  children: (enabled: boolean) => React.ReactNode | null
```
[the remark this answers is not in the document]
cazala: yep, good catch, removed

src/containers/ChainButton/ChainButton.tsx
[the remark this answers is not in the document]
LautaroPetaccio: Oh, that's right, we don't have a CSS processor here.

src/modules/wallet/utils.ts
cazala: The hardcoded value is the default value, but it's set via `createWalletSaga`, here: https://github.com/decentraland/decentraland-dapps/pull/166/files#diff-3653c3a63c9cf3d4372ba903dcd32fbf1778e41d2ce8ce6d4ceaec56a0c5aa03R237

src/modules/wallet/sagas.ts
```diff
@@ -157,6 +171,47 @@ function* handleConnectWalletSuccess() {
...
+    yield call([provider, 'request'], {
+      method: 'wallet_switchEthereumChain',
+      params: [{ chainId: '0x' + chainId.toString(16) }]
+    })
+    yield put(switchNetworkSuccess(chainId))
+  } catch (switchError) {
+    // This error code indicates that the chain has not been added to MetaMask.
+    if (provider && switchError.code === 4902) {
+      try {
+        yield call([provider, 'request'], {
```
LautaroPetaccio, line 190: This is something that I've discovered while reading the tests. As there's a test missing (the failing path of adding an Ethereum chain), the [Metamask docs](https://docs.metamask.io/guide/rpc-api.html#other-rpc-methods) say that the user might choose to switch to the other network after adding the network, but what if they don't? We'll be signaling a `switchNetworkSuccess` but we'll still be on the other network.

LautaroPetaccio (changes requested):

src/modules/wallet/sagas.ts
cazala: good catch! added a check after adding the network to verify the network chainId indeed changed, and also added a test for that case.

src/modules/wallet/selectors.ts
```diff
@@ -28,11 +29,11 @@ export const getNetwork = (state: any) =>
 export const getNetworks = (state: any) =>
   isConnected(state) ? getData(state)!.networks : undefined
 
+// This method is deprecated, it only returns the MANA balance on Ethereum, use getNetworks() to get the MANA balances on all the networks.
```
LautaroPetaccio, line 32: If you add a `@deprecated` JSDOC tag, the method name will be crossed and the deprecation notice will be shown.
[Doc on how to do it](https://jsdoc.app/tags-deprecated.html)

src/modules/wallet/sagas.spec.ts
```diff
@@ -82,12 +82,63 @@ describe('Wallet sagas', async () => {
...
+              [
+                matchers.call.fn(getConnectedProvider),
+                Promise.resolve(mockProvider)
+              ],
+              [
+                call([mockProvider, 'request'], {
+                  method: 'wallet_switchEthereumChain',
+                  params: [{ chainId: '0x1' }]
+                }),
+                Promise.reject(switchError)
```
LautaroPetaccio, line 114: This test is failing on the first switch, but it should return fail later.

LautaroPetaccio (changes requested):

src/modules/wallet/selectors.ts
cazala: done 👍

LautaroPetaccio (approved):

PR #172, opened 2021-08-23
feat: handle unknown networks

cazala:
Added a few helpers to get the network name (either by `Network` or `ChainId`). 

```ts
getNetworkName(network: Network): string
getNetworkNameByChainId(chainId: ChainId): string
```

If the network is not part of our `Network` enum, it will return `Unknown Network (chainId=XX)`. I didn't add this functionality into the `getChainName` helper from `@dcl/schemas` because I didn't want to change the signature, which returns a `ChainName | null`, whereas these helpers return always `string`.

The idea behind this is to avoid the issue described on this comment: https://github.com/decentraland/account/pull/142#pullrequestreview-736478735

### DIFF
--- a/src/containers/ChainCheck/ChainCheck.tsx
+++ b/src/containers/ChainCheck/ChainCheck.tsx
@@ -1,9 +1,11 @@
 import React from 'react'
 import { Popup } from 'decentraland-ui'
 import { T } from '../../modules/translation/utils'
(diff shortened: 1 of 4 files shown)


cazala:
Closed in favor of https://github.com/decentraland/decentraland-dapps/pull/173

PR #173, opened 2021-08-23
feat: handle unknown networks

cazala:
Show a different copy when we don't have a display name for the connected network


LautaroPetaccio (approved):

nicosantangelo (approved):
LGTM

PR #191, opened 2021-11-09
Use NPM `>= 7` instead of  `^7`

2fd:


LautaroPetaccio:
We're actually using NPM `^6`, the engines section is wrong. We're going to migrate to NPM 7 soon.

2fd:
Couldn't we update it to `>=6.0.0` at least?

LautaroPetaccio:
The thing is that we shouldn't say that this is going to work for versions greater or equal than the one we're using it because we'll be allowing major versions with breaking changes that we haven't tested. In this case, we don't want to move to the lockfile v2 yet.

2fd:
cloud we fix the `package.json` to reflex that statement.

```
    engines.npm =  "^6"
```

however if the package supports currents version of npm it should be outdated, even if the team is only using `npm@6`

```
    engines.npm =  "^6 || ^7 || ^8"
```

Please take into account that version 6 and 7 are consider legacy:

Screen Shot 2021-11-22 at 12 53 50

LautaroPetaccio:
As you've mentioned,
```
engines.npm =  "^6"
```
is the correct setting for the moment.

We'll do a full migration once we have all of our pipelines running with the latest LTS node version.

PR #206, opened 2022-01-23
feat: added toasts for meta transaction errors

cazala:
Added a foked saga within the `toastSaga` that listens to meta transaction errors and shows toasts for each error code. If we were to add more error codes (like rate limiting or quota related errors) they can easily be handled within the `handleMetaTransactionError` saga. I also added a `hideAllToasts` action that is triggered when switching networks, to remove the prompt to switch to Polygon. Each `MetaTransactionErrorToast` component supports a custom message (to be customized according to the `ErrorCode`, a button to switch the network to polygon, and an optional `learnMoreLink` prop, which should be a URL to the docs explaining the error and how to workaround it.


jest.config.js
```diff
@@ -1,5 +1,9 @@
 /** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
 module.exports = {
   preset: 'ts-jest',
-  testEnvironment: 'jsdom'
+  testEnvironment: 'jsdom',
+  moduleNameMapper: {
+    '\\.(css|less|sass|scss)$': '<rootDir>/__mocks__/styleMock.js',
+    '\\.(gif|ttf|eot|svg)$': '<rootDir>/__mocks__/fileMock.js'
```
cazala, line 7: these were needed due to an import of `ToastProps` that ended up importing some css files

src/modules/toast/sagas.ts
```diff
@@ -30,3 +50,61 @@ function* handleHideToast(action: HideToastAction) {
...
+
+export function* watchMetaTransactionErrors() {
+  const channel: EventChannel<ErrorCode> = yield call(
+    createMetaTransactionsErrorChannel
+  )
+  while (true) {
+    const code: ErrorCode = yield take(channel)
+    yield call(handleMetaTransactionError, code)
+  }
+}
```
LautaroPetaccio, line 84: I think that they're using the while loop in the example [here](https://redux-saga.js.org/docs/advanced/Channels) to catch for any possible error, but if you're not going to error anywhere, you can just create the channel and use `takeEvery` to listen for the channel. You could also avoid the fork if you do that.

src/modules/transaction/utils.spec.ts
```diff
@@ -194,15 +194,13 @@ describe('modules', function () {
...
+        it('should throw for a normal action (use isTransactionAction to avoid it)', function() {
           const action = {
             type: '[Success] Transaction action',
             payload: { more: 2 }
           }
 
-          expect(() => getTransactionHashFromAction(action)).toThrowError(
-            "Cannot read property 'hash' of undefined"
-          )
+          expect(() => getTransactionHashFromAction(action)).toThrow
```
LautaroPetaccio, line 203: toThrow should be a function call:
```suggestion
          expect(() => getTransactionHashFromAction(action)).toThrow()
```
Was it changed because the error changed?

src/modules/wallet/utils.ts
```diff
@@ -83,57 +84,85 @@ export async function getTargetNetworkProvider(chainId: ChainId) {
...
+    // populate the transaction data
+    const unsignedTx = await getPopulatedTransaction(
+      contractInstance.populateTransaction
+    )
+
+    // if the connected provider is in the target network, use it to sign and send the tx
+    if (chainId === contract.chainId) {
+      const signer = targetNetworkProvider.getSigner()
+      const tx = await signer.sendTransaction(unsignedTx)
+      transactionEvents.emit(TransactionEventType.SUCCESS, { txHash: tx.hash })
```
LautaroPetaccio, line 143: I'm not sure if we should make sendTransaction a sagas handler now that we need to impact the state to show the toasts.

LautaroPetaccio (changes requested):

src/modules/translation/defaults/zh.json
```diff
@@ -81,6 +81,24 @@
       "guest": "来宾",
       "settings": "设定值",
       "account": "帐户"
+    },
+    "敬酒": {
+      "元交易": {
+        "合同帐户错误": {
```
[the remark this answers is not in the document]
cazala: 😅

src/modules/transaction/utils.spec.ts
cazala: Yes, the error message changed between node versions

src/modules/translation/utils.ts
```diff
@@ -51,7 +51,7 @@ export function getCurrentLocale() {
 }
 
 export function t(id: string, values?: any) {
-  return currentLocale.formatMessage({ id }, values)
+  return currentLocale ? currentLocale.formatMessage({ id }, values) : ''
```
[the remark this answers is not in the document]
cazala: The problem here was that the tests would blow up because the TranslationProvider is not present. I made it this way so the tests could run, in practice this shouldn't happen

src/modules/toast/sagas.ts
cazala: mm i don't think i followed how the non-forking part would work, neither why would the loop would be a problem?

LautaroPetaccio (approved):

PR #217, opened 2022-02-20
feat: Make send transaction testable

LautaroPetaccio:
This PR makes the `sendTransaction` function capable of receiving the name of the method to execute and its parameters so it doesn't have to receive an anonymous function, making the function testable.
This change is not a breaking change, the old behavior can be still used.


cazala (approved):

cazala:
awesome job 🚀

PR #231, opened 2022-04-20
feat: Add action to set the profile alias

LautaroPetaccio:
This PR adds a new action to change the profile alias.


nicosantangelo (approved):
lgtm!

PR #236, opened 2022-04-26
feat: Polygon partial support with wallet connect

fzavalia:


src/modules/wallet/utils.ts
```diff
@@ -250,3 +246,20 @@ export function getAddEthereumChainParameters(
       }
   }
 }
+
+export async function getProviderChainId(provider: Provider) {
```
nicosantangelo, line 250: maybe we can add an explicit return type here

src/modules/wallet/utils.ts
```diff
@@ -250,3 +246,20 @@ export function getAddEthereumChainParameters(
...
+    params: []
+  })) as string | number
+
+  let chainId: number
+
+  if (typeof providerChainId === 'string') {
+    chainId = parseInt(providerChainId as string, 16)
+  } else {
+    chainId = providerChainId
+  }
```
nicosantangelo, line 262: Do you think it might be useful to add a comment here explaining what's going on? Or maybe just a JSDoc on the getProviderChainId function

src/modules/wallet/utils.ts
```diff
@@ -17,6 +17,7 @@ import {
 } from '../../lib/eth'
 import { getChainConfiguration } from '../../lib/chainConfiguration'
 import { AddEthereumChainParameters, Networks, Wallet } from './types'
+import { Provider } from 'decentraland-connect/dist'
```
nicosantangelo, line 20: 🔝  !

also is using `/dist` necessary?

nicosantangelo (commented):
👍

fzavalia:
Thanks @nicosantangelo addressed all your comments!

nicosantangelo (approved):

PR #257, opened 2022-06-01
feat: remove web3x

### LINKED ISSUE #222: Reduce the amount of RPC calls when loading the marketplace authorizations
We are checking all the collections approvals when the marketplace loads and therefore we consume a lot of RPC requests.

We need to play a little bit with the [multicall batchSize](https://github.com/0xsequence/sequence.js/tree/master/packages/multicall#configuration) in order to reduce the number of requests. The default (current) is 50.

With a `batchSize` of 1000 the requests consume 8M gas (ok! :check:) and  last 2s.
With a `batchSize` of 500 the requests consume 4M gas (ok :check:) and last 1.5s
With a `batchsize` of 200 the requests consume 1M gas (ok :check:) and last 0.5s

nicosantangelo:
Closes #222


src/modules/transaction/sagas.ts
```diff
@@ -46,6 +45,7 @@ import { isPending, buildActionRef } from './utils'
 import { getTransaction as getTransactionFromChain } from './txUtils'
 import { getAddress } from '../wallet/selectors'
 import { getConnectedProvider } from '../../lib/eth'
+import { providers } from 'ethers'
```
cazala, line 48: 🔝

src/lib/entities.ts
```diff
@@ -1,15 +1,13 @@
...
+import { Entity, EntityType } from '@dcl/schemas'
 import {
   CatalystClient,
   BuildEntityWithoutFilesOptions
 } from 'dcl-catalyst-client'
-import { Entity, EntityType } from 'dcl-catalyst-commons'
 import { getConnectedProvider } from './eth'
 import { ProfileEntity } from './types'
 import { PeerAPI } from './peer'
+import { providers } from 'ethers'
```
cazala, line 10: 🔝

src/providers/WalletProvider/WalletProvider.types.ts
```diff
@@ -1,4 +1,4 @@
-import { EthereumProvider } from 'web3x/providers/ethereum-provider'
+import { providers } from 'ethers'
```
cazala, line 1: should we refactor these to be `ethers.provider` as discussed offline?

cazala (approved):
⛳ .

awesome!  🏌️‍♂️

PR #266, opened 2022-06-20
fix: use appChainId instead of getConnectedProviderChainId

cazala:
There were several places where we used the `getConnectedProviderChainId` to get the chain id that the app expects the user to be. This worked before because that util returned the `connectionData` from `decentraland-connect`, which had a bug were it would not update the `chainId` after changing network, so it kept the value from the first successful connection, and we used that as the app's chain id. Now that the `connetionData` reflects the actual chain id of the user's provider all the time, this breaks the logic from the wallet module sagas and containers. Instead of using that, we need to use the `appChainId` property that's stored on the redux state.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -12,7 +12,7 @@ jobs:
         with:
           node-version: '16'
       - name: Install dependencies
-        run: npm ci --legacy-peer-deps #TODO: remove this flag once we get rig of tslint
+        run: npm ci --legacy-peer-deps #TODO: remove this flag once we get rid of tslint
       - name: Build
         run: npm run build
       - name: Test


src/containers/NetworkCheck/NetworkCheck.tsx
```diff
@@ -1,9 +1,8 @@
 import React from 'react'
 import { Popup } from 'decentraland-ui/dist/components/Popup/Popup'
 import { T } from '../../modules/translation/utils'
-import { getChainIdByNetwork, getConnectedProviderChainId } from '../../lib/eth'
+import { getChainIdByNetwork } from '../../lib/eth'
 import { Props } from './NetworkCheck.types'
-import { getChainName } from '@dcl/schemas/dist/dapps/chain-id'
 import ChainProvider from '../ChainProvider'
 
 export default class ChainCheck extends React.PureComponent<Props> {
```
fzavalia, line 8: Could we rename this to NetworkCheck? I don't think it would be a breaking change because it is exported by default

cazala: done ✅

fzavalia (approved):

PR #267, opened 2022-06-21
fix: getChainIdByNetwork should use appChainId instead of connectedPr…

cazala:
This PR fixes the `getChainIdByNetwork` so it uses the `appChainId` instead of the `connectedProviderChainId`. This used to work until we fixed `decentraland-connect` to always return the connected provider chain id with the `connectionData`.

In this PR I also refactored the `CHAIN_ID` local variable to be used via internal getter/setter methods. I had to do this so I could get the value from a different file/scope.

### DIFF
--- a/src/lib/eth.ts
+++ b/src/lib/eth.ts
@@ -15,6 +15,26 @@ export type EthereumWindow = Window & {
   }
 }
 
+/**
+ * Internal getter/setter to store the app chain id
+ */
+let appChainId: ChainId
+
+/**
+ * @internal
+ */
+export function _getAppChainId() {
+  return appChainId
+}
+
+/**
+ * @internal
+ * @param _appChainId
+ */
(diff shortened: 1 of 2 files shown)


src/lib/eth.ts
```diff
@@ -83,10 +103,10 @@ export function isValidChainId(chainId: string | number) {
 }
 
 export function getChainIdByNetwork(network: Network) {
-  const connectedChainId = getConnectedProviderChainId()
-  if (!connectedChainId) {
-    throw new Error('Could not get connected provider chain id')
+  const appChainId = _getAppChainId()
```
cazala, line 106: This is the fix to the issue with `getChainIdByNetwork` which causes, amog other things, the `NetworkCheck` component to fail to detect if the network is wrong.

fzavalia (approved):
```
  .
 ..^____/
`-. ___ )
  ||   ||
```

PR #271, opened 2022-08-01
feat: move the logic from handleSwitchNetworkRequest sagas into utils

braianj:
We moved part of the chain switch logic used in the handleSwitchNetworkRequest sagas, so that we can use this utility within other projects such as explorer-website.


cazala (approved):
LGTM

PR #292, opened 2022-11-30
fix: Wait for dropped transactions to be completed

LautaroPetaccio:
This PR follows adds to the `waitForTx` util the functionality to wait for dropped transactions that get a different treatment in [the sagas](https://github.com/decentraland/decentraland-dapps/blob/master/src/modules/transaction/sagas.ts).


juanmahidalgo (approved):
looks great 👏

cazala (approved):

PR #301, opened 2022-12-08
feat: add icon to toast provider

meelrossi:
A new property was added to the toast component to accept icons. This pr is to be able to send that icon when calling showToast saga.


cyaiox (approved):
:rocket:

PR #312, opened 2022-12-15
feat: use profile fetched from profile lambda instead of profile entity from catalyst client

### LINKED ISSUE #311: Use name as alias not working as expected for some users
## 🎉 Description

It has been reported that some users are having issues while setting the alias. This ticket is to investigate what's going on and get it fixed.
Hint: seems to be related to how the profile is fetched (instead of getting it from the catalystClient it should be fetched form the /profile lambda)

* [X] 🐛 This is a bug report.
* [ ] 📈 This is a feature request.

juanmahidalgo:
Closes #311

### DIFF
--- a/src/modules/profile/sagas.ts
+++ b/src/modules/profile/sagas.ts
@@ -24,6 +24,7 @@ import {
   setProfileAvatarAliasFailure,
   setProfileAvatarAliasSuccess
 } from './actions'
+import { lambdaProfileToContentProfile } from './utils'
 import { Profile } from './types'
 
 type CreateProfileSagaOptions = {
@@ -115,30 +116,22 @@ export function createProfileSaga({ peerUrl }: CreateProfileSagaOptions) {
     address: string,
     changes: Partial<Avatar>
   ) {
-    const entity: ProfileEntity = yield call(
-      [entities, entities.getProfileEntity],
-      address
-    )
+    const profile: Profile = yield call([peerApi, 'fetchProfile'], address)
+    const profileWithContentHashes = lambdaProfileToContentProfile(profile)
 
     const newAvatar: Avatar = {
-      ...entity.metadata.avatars[0],
+      ...profileWithContentHashes.avatars[0],
       ...changes,
-      version: entity.metadata.avatars[0].version + 1
+      version: profileWithContentHashes.avatars[0].version + 1
     }
 
     const profileMetadata: Profile = {
-      ...entity.metadata,
-      avatars: [newAvatar, ...entity.metadata.avatars.slice(1)]
-    }
-
-    const newProfileEntity: ProfileEntity = {
-      ...entity,
-      metadata: profileMetadata
+      avatars: [newAvatar, ...profileWithContentHashes.avatars.slice(1)]
     }
 
     yield call(
       [entities, 'deployEntityWithoutNewFiles'],
-      newProfileEntity,
+      profileMetadata,
       EntityType.PROFILE,
       address,
       address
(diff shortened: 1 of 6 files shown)


src/modules/profile/sagas.ts
```diff
@@ -115,30 +115,22 @@ export function createProfileSaga({ peerUrl }: CreateProfileSagaOptions) {
...
     const newAvatar: Avatar = {
-      ...entity.metadata.avatars[0],
+      ...profileWithContentHashes.avatars[0],
       ...changes,
-      version: entity.metadata.avatars[0].version + 1
+      version: profileWithContentHashes.avatars[0].version + 1
     }
 
     const profileMetadata: Profile = {
-      ...entity.metadata,
```
LautaroPetaccio, line 130: I would still keep the data from the metadata in case they add something here, so we don't break the profile metadata, what do you think?

src/modules/profile/sagas.ts
```diff
@@ -115,30 +115,22 @@ export function createProfileSaga({ peerUrl }: CreateProfileSagaOptions) {
     address: string,
     changes: Partial<Avatar>
   ) {
-    const entity: ProfileEntity = yield call(
-      [entities, entities.getProfileEntity],
-      address
-    )
+    const profile: Profile = yield call([peerApi, 'fetchProfile'], address)
```
LautaroPetaccio, line 118: `fetchProfile` seems to have some kind of cache. We might want to remove it because it might break the user experience if they want to update more than once their profiles, right?

src/modules/profile/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const lambdaProfileToContentProfile = (profile: Profile) => {
+  return {
+    ...profile,
+    avatars: profile.avatars.map(avatar => ({
+      ...avatar,
+      avatar: {
+        ...avatar.avatar,
+        snapshots: {
+          body: avatar.avatar.snapshots.body.replace(
```
LautaroPetaccio, line 11: This is just a question, but, have we confirmed with platform that those are going to be the only snapshots that will be here? I'm a bit worried that there are more or that they add more of them in the future, making this code break the profiles. Maybe we can add a function here that processes all snapshots and performs the replace.

src/modules/profile/sagas.ts
juanmahidalgo: there's no `metadata` in the profile fetched from the lambda, it only has the `timestamp` and `avatars` object. Ref: https://github.com/decentraland/catalyst/blob/main/lambdas/src/apis/profiles/controllers/profiles.ts#L379-L387C2

src/modules/profile/utils.ts
juanmahidalgo: Yeah, I think I could iterate over the whole object to cover the case they add more data to the snapshots object! will update soon

juanmahidalgo: added!

src/modules/profile/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+      avatar: {
+        ...avatar.avatar,
+        snapshots: Object.entries(avatar.avatar.snapshots)
+          .map(([key, value]) => ({
+            [key]: value.replace(/.*\/content\/contents\//, '')
+          }))
+          .reduce((acc, curr) => {
+            return { ...acc, ...curr }
+          }, {}) as Snapshots
+      }
```
LautaroPetaccio, line 18: `Object.fromEntries` will do the same thing as the reduce:
```suggestion
        snapshots: Object.fromEntries(Object.entries(avatar.avatar.snapshots)
          .map(([key, value]) => ({
            [key]: value.replace(/.*\/content\/contents\//, '')
          })))
      }
```

LautaroPetaccio (approved):

Issue #313, opened 2022-12-16
Bypass cache when getting the profile on the "set alias" flow


juanmahidalgo:
We have a cache mechanism in place, which stores the profiles that are fetched. Once deployed with a new alias, the cache is not updated, so it could happen that the user goes to the explorer and makes a change, which causes the next profile deploy to override the other changes. This is an edge case but avoiding the cache mecanism for the aliases would solve the issue.

PR #314, opened 2022-12-16
feat: add a flag to bypass profile cache

juanmahidalgo:
#313

### DIFF
--- a/src/lib/peer.spec.ts
+++ b/src/lib/peer.spec.ts
@@ -57,6 +57,9 @@ describe('peerAPI', () => {
           expect(await peerApi.fetchProfile(address, { useCache: true })).toBe(
             profileUpdated
           )
+          await expect(peerApi.cache[address]).resolves.toStrictEqual(
+            profileUpdated
+          )
         })
       })
     })
@@ -83,6 +86,9 @@ describe('peerAPI', () => {
       })
       it('should fetch the profile', async () => {
         expect(await peerApi.fetchProfile(address, { useCache: false })).toBe(
+          profileUpdated
+        )
+        await expect(peerApi.cache[address]).resolves.toStrictEqual(
           profileUpdated
         )
       })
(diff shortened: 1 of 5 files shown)


src/lib/types.ts
```diff
@@ -38,3 +38,7 @@ export type StorageOwnData = {
     version: number
   }
 }
+
+export type FetchProfileOptions = {
+  useCache: boolean
```
LautaroPetaccio, line 43: As it is a set of options, should we mark them as optional as well?
```suggestion
  useCache?: boolean
```

src/lib/peer.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+          } as Profile
+          peerApi.cache = {}
+          jest
+            .spyOn(peerApi.lambdasClient, 'fetchProfiles')
+            .mockImplementation(() => Promise.resolve([profileUpdated]))
+        })
+        it('should fetch the profile', async () => {
+          expect(await peerApi.fetchProfile(address, { useCache: true })).toBe(
+            profileUpdated
+          )
```
LautaroPetaccio, line 59: Is it possible to check that the cache was updated as well? Like, for example, we might be able to perform an:
```
await expect(peerApi.cache).resolves.toEqual(profileUpdated)
```
What if we add something like this to ensure the cache to be set?
The same applies for the case when the `useCache` parameter is not passed.

LautaroPetaccio (changes requested):
Great job 👍

src/lib/types.ts
juanmahidalgo: updated!

src/lib/peer.spec.ts
juanmahidalgo: added! in the case the `useCache` parameter is not used, the cache is set anyways, so the `expect` is the same in both cases.

LautaroPetaccio (approved):

LautaroPetaccio:
Goes out with 🍟 , well done!

PR #324, opened 2022-12-22
feat: add "hashesByKey" when building profile entity before deploying it

### LINKED ISSUE #323: Send the `hashesByKey` field when deploying profiles
As a new requirement to deploy profiles that were built with the `buildEntityWithoutNewFiles` function from the catalystClient, we need to send the `hashesByKey` as well. It will have the information of the snapshots, translated into a `Map` with the `name` as `face256` and `body` and the key being the content hash of the file content.

juanmahidalgo:
Closes #323

### DIFF
--- a/src/lib/peer.ts
+++ b/src/lib/peer.ts
@@ -27,7 +27,9 @@ export class PeerAPI extends BaseAPI {
     }
 
     this.cache[address] = this.lambdasClient
-      .fetchProfiles([address.toLowerCase()])
+      .fetchProfiles([address.toLowerCase()], undefined, {
+        headers: { 'Cache-Control': 'max-age=0' }
+      })
       .then(profiles => profiles[0])
 
     return this.cache[address]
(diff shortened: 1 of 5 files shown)


cazala (approved):

src/lib/peer.ts
```diff
@@ -27,7 +27,9 @@ export class PeerAPI extends BaseAPI {
     }
 
     this.cache[address] = this.lambdasClient
-      .fetchProfiles([address.toLowerCase()])
+      .fetchProfiles([address.toLowerCase()], undefined, {
+        headers: { 'Cache-Control': 'max-age=0' }
```
menduz, line 31: @juanmahidalgo you may also want to add `no-cache`

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.32

juanmahidalgo: sure, will do it in the following PR!

PR #325, opened 2022-12-22
feat: Add in-progress and failure states to the Buy MANA flow

### LINKED ISSUE #315: Convert BuyManaWithFiatModal into a modal usable by the ModalProvider
It will save us from implementing the close modal in each project that will use the modal.

It will need to receive by props, at least, `onClose` and `metadata`.

### LINKED ISSUE #321: Add in-progress and failure states to the Buy MANA flow
Blocked by

kevinszuchet:
Closes #321 
Closes #315

### DIFF
--- a/src/containers/BuyManaWithFiatModal/BuyManaWithFiatFeedbackModal/BuyManaWithFiatFeedbackModal.tsx
+++ b/src/containers/BuyManaWithFiatModal/BuyManaWithFiatFeedbackModal/BuyManaWithFiatFeedbackModal.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useCallback } from 'react'
 import {
   FeedbackModal as BaseBuyManaWithFiatFeedbackModal,
   FeedbackModalI18N,
@@ -66,7 +66,7 @@ const BuyManaWithFiatFeedbackModal = ({
   const transactionStatus = transactionStatuses[purchaseStatus]
   const analytics = getAnalytics()
 
-  const handleCtaClick = () => {
+  const handleCtaClick = useCallback(() => {
     switch (transactionStatus) {
       case TransactionStatus.SUCCESS:
         onClose()
@@ -81,9 +81,9 @@ const BuyManaWithFiatFeedbackModal = ({
       default:
         break
     }
-  }
+  }, [transactionStatus, analytics, network, gateway, onClose, onTryAgain])
 
-  const handleSecondaryCtaClick = () => {
+  const handleSecondaryCtaClick = useCallback(() => {
     if (transactionStatus === TransactionStatus.FAILURE) {
       analytics.track('Select other gateway', {
         network,
@@ -92,7 +92,14 @@ const BuyManaWithFiatFeedbackModal = ({
       onSelectOtherProvider(network)
       onClose()
     }
-  }
+  }, [
+    transactionStatus,
+    analytics,
+    network,
+    gateway,
+    onSelectOtherProvider,
+    onClose
+  ])
 
   return (
     <BaseBuyManaWithFiatFeedbackModal
--- a/src/containers/BuyManaWithFiatModal/BuyManaWithFiatFeedbackModal/BuyManaWithFiatFeedbackModal.tsx
+++ b/src/containers/BuyManaWithFiatModal/BuyManaWithFiatFeedbackModal/BuyManaWithFiatFeedbackModal.tsx
@@ -1,34 +1,99 @@
 import React from 'react'
-import { Network } from '@dcl/schemas'
 import {
   FeedbackModal as BaseBuyManaWithFiatFeedbackModal,
-  FeedbackModalI18N
+  FeedbackModalI18N,
+  TransactionStatus
 } from 'decentraland-ui/dist/components/BuyManaWithFiatModal/FeedbackModal'
+import {
+  gatewaysNames,
+  networksNames
+} from 'decentraland-ui/dist/components/BuyManaWithFiatModal/Network'
+import { getAnalytics } from '../../../modules/analytics/utils'
+import { Purchase, PurchaseStatus } from '../../../modules/mana/types'
 import { t } from '../../../modules/translation/utils'
 import { Props } from './BuyManaWithFiatFeedbackModal.types'
 
+const transactionStatuses = {
+  [PurchaseStatus.PENDING]: TransactionStatus.PENDING,
+  [PurchaseStatus.COMPLETE]: TransactionStatus.SUCCESS,
+  [PurchaseStatus.FAILED]: TransactionStatus.FAILURE,
+  [PurchaseStatus.CANCELLED]: TransactionStatus.FAILURE
+}
+
+// TODO: fix the errors raised because of missing translations
 const getDefaultFeedbackTranslations = (
-  network: Network
-): FeedbackModalI18N => ({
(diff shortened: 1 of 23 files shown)


src/modules/manaFiatGateway/selectors.ts
```diff
@@ -8,9 +9,15 @@ import { ManaFiatGatewayState } from './reducer'
 export const getState: (state: any) => ManaFiatGatewayState = state =>
   state.manaFiatGateway
 
+export const getData = (state: any) => getState(state).data
 export const getLoading = (state: any) => getState(state).loading
 export const getError = (state: any) => getState(state).error
 
+export const getPendingPurchase = (state: any) =>
+  getData(state).purchases?.find(
```
juanmahidalgo, line 17:
```suggestion
  getData(state).purchases.find(
```
I don't think the `?` is needed here since it's already initialized as an empty array, it should never be `undefined`

src/modules/manaFiatGateway/selectors.ts
```diff
@@ -8,9 +9,15 @@ import { ManaFiatGatewayState } from './reducer'
 export const getState: (state: any) => ManaFiatGatewayState = state =>
   state.manaFiatGateway
 
+export const getData = (state: any) => getState(state).data
 export const getLoading = (state: any) => getState(state).loading
 export const getError = (state: any) => getState(state).error
 
+export const getPendingPurchase = (state: any) =>
```
juanmahidalgo, line 16:
```suggestion
export const getPendingPurchase = (state: ManaFiatGatewayState) =>
```
I know the other ones were already there and they all have `any`, but I believe we could start typing it as `ManaFiatGatewayState`, right?

src/containers/BuyManaWithFiatModal/BuyManaWithFiatFeedbackModal/BuyManaWithFiatFeedbackModal.tsx
```diff
@@ -1,34 +1,110 @@
...
+  metadata: { purchase, goToUrl },
+  onTryAgain,
+  onSelectOtherProvider,
   onClose
 }: Props) => {
+  const { network, gateway, status: purchaseStatus } = purchase
+  const transactionStatus = transactionStatuses[purchaseStatus]
+  const analytics = getAnalytics()
+
+  const handleCtaClick = () => {
```
juanmahidalgo, line 69: nit pick: could we wrap this with `useCallback` so we avoid re-definiting them on each render?  Same for the `handleSecondaryCtaClick` one

src/containers/BuyManaWithFiatModal/BuyManaWithFiatFeedbackModal/BuyManaWithFiatFeedbackModal.tsx
```diff
@@ -1,34 +1,110 @@
...
+      onSelectOtherProvider(network)
+      onClose()
+    }
+  }
+
   return (
     <BaseBuyManaWithFiatFeedbackModal
       open
+      status={transactionStatus}
+      selectedNetwork={purchase.network}
```
juanmahidalgo, line 101: Now it might be a little bit too late, but I think it could have been cleaner to send the whole `purchase` as prop and the modal could have read the fields from it (network, gateway, tx status, etc). Just a comment in case you wanna refactor it later (not needed for this PR)

src/modules/location/sagas.ts
```diff
@@ -12,15 +13,22 @@ export function* locationSaga() {
...
-  } = query as RedirectTransactionQuery
+  } = location.query as RedirectTransactionQuery
 
   if (transactionId && transactionStatus && network && gateway) {
+    const queryParams = new URLSearchParams(location.search)
+    const params = ['transactionId', 'transactionStatus', 'network', 'gateway']
+
+    params.forEach(param => queryParams.delete(param))
+
+    yield put(replace(`${location.pathname}?${queryParams.toString()}`))
```
juanmahidalgo, line 30: can we add a little test for this 🙏 ?

src/modules/manaFiatGateway/selectors.ts
kevinszuchet: I think it should be `RootState` instead of the  `ManaFiatGatewayState`. The second type is what you get after getting the `manaFiatGateway` from the `state: RootState`.

juanmahidalgo: Oh yeah, `RootState`, sorry, that's right!

src/containers/BuyManaWithFiatModal/BuyManaWithFiatFeedbackModal/BuyManaWithFiatFeedbackModal.tsx
kevinszuchet: It was a conscious decision, maybe not the best. The UI doesn't understand purchases, so I thought about mapping the purchase status, for example, into the transaction status that the modal understands.

src/containers/BuyManaWithFiatModal/BuyManaWithFiatFeedbackModal/BuyManaWithFiatFeedbackModal.tsx
```diff
@@ -1,34 +1,117 @@
...
+): FeedbackModalI18N => {
+  const basePath = `@dapps.buyManaWithFiat.feedback_modal.${status}`
+  return Object.fromEntries(
+    propsToTranslateByStatus[status].map(it => [
+      it,
+      t(`${basePath}.${camelToSnakeCase(it)}`, {
+        network: networksNames[network],
+        gateway: gatewaysNames[gateway]
+      })
+    ])
```
fzavalia, line 55:
```suggestion
    propsToTranslateByStatus[status].map(prop => [
      prop,
      t(`${basePath}.${camelToSnakeCase(prop)}`, {
        network: networksNames[network],
        gateway: gatewaysNames[gateway]
      })
    ])
```

fzavalia: A more signification name

src/containers/BuyManaWithFiatModal/BuyManaWithFiatModal.container.ts
```diff
@@ -30,8 +30,8 @@ const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
 const mergeProps = (
```
fzavalia, line 30: We could remove this merge props as it doesn't seem to be doing anything

src/containers/BuyManaWithFiatModal/BuyManaWithFiatModal.tsx
```diff
@@ -122,42 +122,33 @@ export default class BuyManaWithFiatModal extends React.PureComponent<
   handleOnContinue(network: Network, gateway: NetworkGatewayType) {
     this.analytics.track('Choose Gateway', { network, gateway })
     this.props.onContinue?.(network, gateway)
```
fzavalia, line 124: Maybe call the analytics when onContinue is set so no false tracks are made.

fzavalia (changes requested):
Good job, some little things I've seen.

fzavalia (approved):

PR #336, opened 2023-01-03
feat: add no-cache header

juanmahidalgo:
As suggested [here](https://github.com/decentraland/decentraland-dapps/pull/324#discussion_r1060045835), this adds the `Pragma: no-cache` header.

### DIFF
--- a/src/lib/peer.ts
+++ b/src/lib/peer.ts
@@ -28,7 +28,7 @@ export class PeerAPI extends BaseAPI {
 
     this.cache[address] = this.lambdasClient
       .fetchProfiles([address.toLowerCase()], undefined, {
-        headers: { 'Cache-Control': 'max-age=0' }
+        headers: { 'Cache-Control': 'max-age=0', Pragma: 'no-cache' }
       })
       .then(profiles => profiles[0])
 


src/lib/peer.ts
```diff
@@ -28,7 +28,7 @@ export class PeerAPI extends BaseAPI {
 
     this.cache[address] = this.lambdasClient
       .fetchProfiles([address.toLowerCase()], undefined, {
-        headers: { 'Cache-Control': 'max-age=0' }
+        headers: { 'Cache-Control': 'max-age=0', Pragma: 'no-cache' }
```
juanmahidalgo, line 31: @menduz just double-checking here, the one you suggested was the `Pragma` one, right? Is it due to the fact that the server might not be HTTP/1.1 compliant?

menduz: It may be honored by servers, although you can also change or add the max-age=0 by no-cache to signal the servers and browser that you don't desire a cached version. But I don't know exactly if max-age works in client requests

juanmahidalgo: [Seems like](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9.4) the `max-age=0` will re-validate the cached version all the way to the server, and if it returns a 304, the cached version will be used (and the body will be ignored). I believe that's a little bit more performant than the `no-cache` and will always get us the latest profile as well (if the cached version is not longer valid, it will get the latest from the server). What do you think? 

note: the rfc states: 

> If a request includes the no-cache directive, it SHOULD NOT include min-fresh, max-stale, or max-age.

so I think we gotta choose between the `Cache-control: 'no-cache'` or  `Cache-control: 'max-age=0'` and not both.

menduz: Thanks for the info! it is your call!

kevinszuchet (approved):

PR #338, opened 2023-01-05
fix: Remove the Pragma no-cache header

juanmahidalgo:


fzavalia (approved):

src/lib/peer.ts
```diff
@@ -28,7 +28,7 @@ export class PeerAPI extends BaseAPI {
 
     this.cache[address] = this.lambdasClient
       .fetchProfiles([address.toLowerCase()], undefined, {
-        headers: { 'Cache-Control': 'max-age=0', Pragma: 'no-cache' }
+        headers: { 'Cache-Control': 'max-age=0' }
```
juanmahidalgo, line 31: FYI: The catalyst are not supporting the `Pragma` header 😢 so I'm removing it for now.
image

juanmahidalgo: @menduz FYI

PR #343, opened 2023-01-24
feat: Extends Transak for supporting nfts purchases + manaFiatGateway to gateway module

kevinszuchet:
Related to https://github.com/decentraland/marketplace/pull/1344


src/modules/gateway/sagas.ts
```diff
@@ -39,16 +38,16 @@ import {
...
-import { ManaFiatGatewaySagasConfig } from './types'
+import { ManaFiatGatewaySagasConfig, Purchase, PurchaseStatus } from './types'
 import { purchaseEventsChannel } from './utils'
 
 const DEFAULT_POLLING_DELAY = 3000
 const BUY_MANA_WITH_FIAT_FEEDBACK_MODAL_NAME = 'BuyManaWithFiatFeedbackModal'
 
-export function createManaFiatGatewaysSaga(config: ManaFiatGatewaySagasConfig) {
-  return function* manaFiatGatewaysSaga(): IterableIterator<ForkEffect> {
+export function createGatewaySaga(config: ManaFiatGatewaySagasConfig) {
```
fzavalia, line 49: Rename to `createGatewaysSaga` (Was in plural before)

src/modules/gateway/sagas.ts
```diff
@@ -248,38 +247,40 @@ export function* handleSetPurchase(action: SetPurchaseAction) {
...
 
-  if (
-    purchase.gateway === NetworkGatewayType.TRANSAK &&
-    purchase.status === PurchaseStatus.PENDING
-  ) {
-    window.addEventListener('beforeunload', handleBeforeUnload)
-  } else {
-    window.removeEventListener('beforeunload', handleBeforeUnload)
-  }
+  if (!nft) {
```
fzavalia, line 252: Maybe this can be refactored to:

```
if (!nft) {
  yield call(handleManaSetPurchase, purchase)
}
```

(Pick a name for the handler that might be a better fit)

src/modules/gateway/selectors.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+    contractAddress: mockContractAddress,
+    tokenId: mockTokenId,
+    tradeType: TradeType.PRIMARY,
+    cryptoAmount: 100
+  }
+}
+
+describe('MANA-FIAT Gateway selectors', () => {
+  beforeEach(() => {
+    manaFiatGatewayState = { gateway: INITIAL_STATE }
```
fzavalia, line 52: This would be the application state, not just the manaFiatGatewayState.

Also, manaFiatGatewayState should be renamed to just gatewayState.

src/modules/gateway/selectors.ts
```diff
@@ -1,24 +1,38 @@
...
 
 export const getPurchases = (state: any) => getData(state).purchases
-export const getPendingPurchase = (state: any) =>
+export const getPendingManaPurchase = (state: any) =>
   getPurchases(state).find(
-    purchase => purchase.status === PurchaseStatus.PENDING
+    purchase => purchase.status === PurchaseStatus.PENDING && !purchase.nft
   )
 
+export const getNFTPurchase = (
```
fzavalia, line 24: Add a comment of why the purchases have to be sorted.

src/modules/gateway/selectors.ts
```diff
@@ -1,24 +1,38 @@
...
+  state: any,
+  contractAddress: string,
+  tokenId: string
+) =>
+  getPurchases(state)
+    .sort(sortByTimestamp)
```
fzavalia, line 30: Sort will modify the original array. 

Store it first in a different array,

```
const arr = [...getPurchases(state)]
```

and then sort `arr`

src/modules/gateway/types.ts
```diff
@@ -0,0 +1,49 @@
...
+type NFTPurchase = {
+  nft?: {
+    contractAddress: string
+    tokenId: string
+    tradeType: TradeType
+    cryptoAmount: number
+  } | null
+}
+
+export type Purchase = ManaPurchase & NFTPurchase
```
fzavalia, line 49: I think the correct way to type this is:
```
BasePurchase = {
  id: string
  network: Network
  gateway: NetworkGatewayType
  timestamp: number
  status: PurchaseStatus
  address: string
  txHash: string | null
}

ManaPurchase = BasePurchase & { amount: number }

NFTPurchase = BasePurchase & { nft: {...} }

export type Purchase = ManaPurchase || NFTPurchase
```

fzavalia (changes requested):

src/modules/gateway/sagas.ts
kevinszuchet: It was a typo. It should be `create{Name of the module}Saga`. In this case the module is `gateway`.

src/modules/gateway/selectors.ts
kevinszuchet: Nice catch 💯

src/modules/gateway/sagas.ts
fzavalia: But then it creates the `gatewaysSaga` 🤔

src/modules/gateway/transak/Transak.ts
```diff
@@ -16,29 +15,15 @@ const PURCHASE_EVENT = 'Purchase status change'
...
-      hostURL: window.location.origin,
-      widgetHeight: '650px',
-      widgetWidth: '450px'
-    }
-  }
-
-  protected customizationOptions(address: string): CustomizationOptions {
-    return {
-      ...this.defaultCustomizationOptions(address),
+    this.customizationOptions = customizationOptions || {
```
fzavalia, line 26: If customizationOptions is provided 

```
defaultCryptoCurrency: 'MANA',
cyptoCurrencyList: 'MANA',
fiatCurrency: '', // INR/GBP
```

Will not be set.

I recommend just setting an empty object on the constructor and adding those 3 to the defautlOptions

src/modules/gateway/utils.ts
```diff
@@ -0,0 +1,8 @@
+import { channel } from 'redux-saga'
+import { ManaPurchase, Purchase } from './types'
+
+export const purchaseEventsChannel = channel()
+
+export function isManaPurchase(purchase: Purchase): purchase is ManaPurchase {
+  return 'amount' in purchase && !('nft' in purchase)
```
fzavalia, line 7: Just by checking that it doesn't have nft is enough

src/modules/gateway/sagas.ts
```diff
@@ -241,47 +241,52 @@ const handleBeforeUnload = (e: BeforeUnloadEvent) => {
...
+        transactionUrl
+      })
+    )
+  }
+}
+
+export function* handleSetPurchase(action: SetPurchaseAction) {
+  const { purchase } = action.payload
+
+  if (!purchase.nft) {
```
fzavalia: Didn't you make a util to check this?

fzavalia (changes requested):

src/modules/gateway/sagas.ts
fzavalia: `isManaPurchase `

src/modules/gateway/transak/Transak.ts
kevinszuchet: They are not default. They are attached to the MANA purchases.

fzavalia: They should be provided via constructor when doing mana purchases don't you think?

fzavalia (approved):

PR #345, opened 2023-01-25
fix: update @dcl/schemas and related dependencies

2fd:
This update to the dependencies and import changes helps to reduce the bundle size on projects that depends on this repo

| before | after |
| ------ | ----- |
| before | after |

Issues:

- decentraland/unity-renderer#4027


src/lib/entities.ts
```diff
@@ -57,7 +55,6 @@ export class EntitiesOperator {
    */
   async getProfileEntity(address: string): Promise<ProfileEntity> {
     const entities: Entity[] = await this.catalystClient.fetchEntitiesByPointers(
-      EntityType.PROFILE,
```
cyaiox, line 60: is this property no longer required, or is the method `fetchEntitiesByPointers` getting the `type` as `EntityType.PROFILE` by default?

src/lib/peer.ts
```diff
@@ -26,8 +26,10 @@ export class PeerAPI extends BaseAPI {
       return this.cache[address]
     }
 
+    this.lambdasClient.fetchProfiles
```
cyaiox, line 29: should be called this method?

2fd: my bad

src/lib/entities.ts
2fd: the new version of catalystClient removed that prop because wasn't necessary at all, all entities have unique ids, no matter the type

cyaiox (approved):
LGTM! 🚀

PR #348, opened 2023-01-26
refactor: modules/wallet/utils

2fd:
This PR breaks `modules/wallet/utils` into multiple files; this way dependent projects can import only the code they need


src/modules/wallet/utils/urls.ts
```diff
@@ -0,0 +1,4 @@
+let TRANSACTIONS_API_URL = 'https://transactions-api.decentraland.co/v1'
```
menduz, line 1:
```suggestion
let TRANSACTIONS_API_URL = 'https://transactions-api.decentraland.org/v1'
```

.co is an internal domain

cazala (approved):

PR #370, opened 2023-02-14
feat: Recover Transak order status with polling

### LINKED ISSUE #332: [Transak] Purchases with pending status forever
1. Open Buy MANA Modal
2. Select Transak
3. Buy some MANA
4. Close the tab while the purchase is still in progress

When you open a new tab, you will see the modal with status pending forever because the process to check the order status was stopped when the user closed the tab.

### LINKED ISSUE #366: Add Payment Method to the Purchase type

kevinszuchet:
Closes #332 
Closes #366

### DIFF
--- a/src/modules/gateway/sagas.spec.ts
+++ b/src/modules/gateway/sagas.spec.ts
@@ -22,10 +22,19 @@ import {
   openManaFiatGatewaySuccess
 } from './actions'
 import { MoonPay } from './moonpay'
-import { MoonPayTransaction, MoonPayTransactionStatus } from './moonpay/types'
+import {
+  MoonPayPaymentMethod,
+  MoonPayTransaction,
+  MoonPayTransactionStatus
+} from './moonpay/types'
 import { createGatewaySaga } from './sagas'
 import { Transak } from './transak'
-import { ManaFiatGatewaySagasConfig, Purchase, PurchaseStatus } from './types'
+import {
+  ManaFiatGatewaySagasConfig,
+  Purchase,
+  PurchasePaymentMethod,
+  PurchaseStatus
+} from './types'
 import { getPendingManaPurchase } from './selectors'
 
 jest.mock('../../lib/eth')
@@ -91,7 +100,7 @@ const mockTransaction: MoonPayTransaction = {
   usdRate: 1.11336,
   gbpRate: 0.86044,
   externalTransactionId: '2dad9a78-bc35-452b-aa2c-968a247a9646',
-  paymentMethod: 'credit_debit_card',
+  paymentMethod: MoonPayPaymentMethod.CREDIT_DEBIT_CARD,
   baseCurrency: {
     id: '71435a8d-211c-4664-a59e-2a5361a6c5a7',
     createdAt: '2022-02-28T11:17:08.116Z',
@@ -179,6 +188,7 @@ const mockPurchase: Purchase = {
   network: Network.ETHEREUM,
   timestamp: 1535398843748,
   status: PurchaseStatus.PENDING,
+  paymentMethod: PurchasePaymentMethod.CREDIT_DEBIT_CARD,
   gateway: NetworkGatewayType.MOON_PAY,
   txHash: null
 }
--- a/src/modules/gateway/sagas.spec.ts
+++ b/src/modules/gateway/sagas.spec.ts
@@ -27,19 +27,10 @@ import {
   openManaFiatGatewaySuccess
 } from './actions'
 import { MoonPay } from './moonpay'
-import {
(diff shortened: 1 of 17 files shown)


src/modules/gateway/sagas.spec.ts
```diff
@@ -22,14 +27,24 @@ import {
...
+  Purchase,
+  PurchasePaymentMethod,
+  PurchaseStatus
+} from './types'
+import { getPendingPurchase } from './selectors'
+import { OrderResponse, TransakOrderStatus } from './transak/types'
 
 jest.mock('../../lib/eth')
-jest.mock('./transak')
+// jest.mock('./transak')
```
fzavalia, line 47: Remove

fzavalia (changes requested):

src/modules/gateway/types.ts
```diff
@@ -30,14 +32,27 @@ export enum PurchaseStatus {
...
 
+export enum PurchasePaymentMethod {
+  CREDIT_DEBIT_CARD = 'credit_debit_card',
+  BANK_TRANSFER = 'bank_transfer',
+  APPLE_PAY = 'apple_pay',
+  GOOGLE_PAY = 'google_pay',
+  // TODO (buy MANA & NFTs with FIAT): finish the mapping between payment methods
+  // https://docs.moonpay.com/moonpay/apis-and-webhooks/apis/client-side-apis/transactions
+  // https://docs.transak.com/docs/control-the-fiat-or-crypto-options-for-the-user#payment-method
+  OTHER = 'other'
```
kevinszuchet, line 43: How should we map the rest of the methods? @fzavalia

src/modules/gateway/sagas.ts
```diff
@@ -157,26 +166,75 @@ function* upsertPurchase(
...
+  if (pendingPurchases) {
+    for (const pendingPurchase of pendingPurchases) {
+      const { network, gateway, id } = pendingPurchase
+      switch (gateway) {
+        case NetworkGatewayType.TRANSAK:
+          yield put(pollPurchaseStatusRequest(pendingPurchase))
+          break
+        case NetworkGatewayType.MOON_PAY:
+          yield put(
+            manaFiatGatewayPurchaseCompleted(
```
fzavalia, line 182: Is this necessary? isn't it just updating the purchase to its same state?

src/modules/gateway/sagas.ts
```diff
@@ -157,26 +166,75 @@ function* upsertPurchase(
...
+      }
+    }
+  }
+}
+
+function* handlePollPurchaseStatusRequest(
+  config: ManaFiatGatewaySagasConfig,
+  action: PollPurchaseStatusRequestAction
+) {
+  const { purchase } = action.payload
```
fzavalia, line 199: This saga should only accept pending purchases, if by some reason it receives a completed one, it will poll forever

src/modules/gateway/transak/types.ts
```diff
@@ -97,6 +97,19 @@ export type OrderData = {
   }
 }
 
+export type OrderResponse = {
+  meta: {
+    orderId: string
+    apiKey: string
+  }
+  data: {
```
fzavalia, line 105: You can update this to 
```
Pick<OrderData['status'], 'id' | 'status' | 'transactionHash'> & {errorMessage?: string | null}
```
If you want to make it fancier

fzavalia: The error message being null and undefined is kind of weird

src/modules/gateway/sagas.ts
kevinszuchet: The default behavior of our MoonPay integration is to poll the order status until it reaches a final one. In this case, it will dispatch the `manaFiatGatewayPurchaseCompleted` action to use that same logic. It will start as a pending order and change to different statuses after fetching that information from MoonPay API.

src/modules/gateway/sagas.ts
```diff
@@ -200,37 +200,39 @@ function* handlePollPurchaseStatusRequest(
...
-          if (newStatus !== purchase.status) {
-            statusHasChanged = true
-            yield put(
-              setPurchase({
-                ...purchase,
-                status: newStatus,
-                txHash: transactionHash || null,
-                failureReason: errorMessage
-              })
+    if (purchase.status === PurchaseStatus.PENDING) {
```
fzavalia, line 203: In this cases its better to just 
```
if (something) {
  return
}

somethingElse code
```

Instead of wrapping the whole code on an if

fzavalia (changes requested):

fzavalia (approved):

PR #371, opened 2023-02-14
feat: sign in modal with metamask

flobarreto:
missing update of libraries: dcl/schemas, decentraland-connect and decentraland-ui


src/containers/LoginModal/utils.ts
```diff
@@ -46,6 +49,8 @@ export function toProviderType(
   modalOptionType: LoginModalOptionType
 ): ProviderType {
   switch (modalOptionType) {
+    case METAMASK_MOBILE:
```
juanmahidalgo, line 52: let's add a comment here saying that we're adding METAMASK_MOBILE with the same WALLET_CONNECT logic, so if we come back later to this code we know why we put it like this :)

juanmahidalgo (approved):
LGTM!

PR #372, opened 2023-02-15
feat: Make partially supported network modal optional

LautaroPetaccio:
This PR makes the "partially supported network modal" optional so we can choose to not show it if we need to change the user experience.


src/containers/Navbar/Navbar.types.tsx
```diff
@@ -16,6 +16,7 @@ export type NavbarProps = NavbarComponentProps & {
   onSwitchNetwork: typeof switchNetworkRequest
   onSignOut: typeof disconnectWallet
   hasAcceptedNetworkPartialSupport: boolean
+  showPartiallySupportedModal: boolean
```
juanmahidalgo, line 19:
```suggestion
  showPartiallySupportedModal?: boolean
```
right?

juanmahidalgo (approved):
looks great 👍

PR #374, opened 2023-02-17
feat: DAPPS Application Name

fzavalia:
Changes:
- Add DAPPS app name to enum
- `StateWithFeatures` type has a possibly undefined features property. Clients might not be implementing the reducer so it might be possible that it is undefined. By typing it accordingly, we might be able to handle it right.
- `getState` now throws a `"'features' module not implemented"` error.
- `getIsFeatureEnabled` returns false if the provided app cannot be found in the state. There might be cases where the features are still loading and the app hasn't been added.

### DIFF
--- a/src/modules/features/selectors.ts
+++ b/src/modules/features/selectors.ts
@@ -66,7 +66,7 @@ export const getIsFeatureEnabled = (
     return false
   }
 
-  return !!appFlags.flags[`${app}-${feature}`] ?? false
+  return !!appFlags.flags[`${app}-${feature}`]
 }
 
 export const isLoadingFeatureFlags = (state: StateWithFeatures) => {
--- a/src/modules/features/selectors.ts
+++ b/src/modules/features/selectors.ts
@@ -11,10 +11,10 @@ import {
 export const getState = (state: StateWithFeatures): FeaturesState => {
   const { features } = state
 
-  // If the features prop is undefined in the state, it is because the client has not implemented the features module.
-  // If any feature inside this lib requires querying the features module, it should try-catch any query to the state and handle accordingly. (Ignore checking for feature flags, etc.)
+  // If the features prop is undefined in the state, it is because the client has not implemented the features reducer.
+  // If any feature inside this lib requires querying the features reducer, it should try-catch any query to the state and handle accordingly. (Ignore checking for feature flags, etc.)
   if (!features) {
-    throw new Error("'features' module not implemented")
+    throw new Error("'features' reducer not implemented")
   }
 
   return features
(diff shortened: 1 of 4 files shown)


src/modules/features/types.ts
```diff
@@ -38,5 +39,7 @@ export type FeatureSagasConfig = {
 }
 
 export type StateWithFeatures = {
-  features: FeaturesState
+  // Possibly undefined because clients might not have implemented te features module into their dapps.
```
cyaiox, line 42: typo error.
`// Possibly undefined because clients might not have implemented the features module into their dapps.`

fzavalia: Thanks 😄

src/modules/features/selectors.ts
```diff
@@ -41,28 +50,31 @@ export const getIsFeatureEnabled = (
...
-  if (!appFeatures) {
-    throw new Error(`Application "${app}" not found`)
+  // The app might not be defined in the store because the flags might not have been fetched yet.
+  // We suggest using isLoadingFeatureFlags and hasLoadedInitialFlags to handle this first.
+  if (!appFlags) {
+    return false
   }
 
-  return !!appFeatures.flags[`${app}-${feature}`]
+  return !!appFlags.flags[`${app}-${feature}`] ?? false
```
cyaiox, line 69: the double not operator will return `false` if the value is `null || undefined || false`

fzavalia: That is the expected result

juanmahidalgo, line 69:
```suggestion
  return !!appFlags.flags[`${app}-${feature}`]
```
`!!undefined` would return `false`, right? if so, I think this `??` is redundant

src/modules/features/selectors.ts
```diff
@@ -41,28 +50,31 @@ export const getIsFeatureEnabled = (
...
-  const features = getData(state)
-  const appFeatures: ApplicationFeatures | undefined = features[app]
+  const appFlags = getData(state)[app]
 
-  if (!appFeatures) {
-    throw new Error(`Application "${app}" not found`)
+  // The app might not be defined in the store because the flags might not have been fetched yet.
+  // We suggest using isLoadingFeatureFlags and hasLoadedInitialFlags to handle this first.
+  if (!appFlags) {
+    return false
```
juanmahidalgo, line 66: quick question! what's the motivation behind this change? returning `false` instead of throwing. Is there any scenario were we prefer that?

src/modules/features/selectors.ts
fzavalia: Ohh you are right, was trying to see if `?? false` was better than `!!` and ended up using both 🙃 Will remove one. @cyaiox that's probably what you meant as well.

cyaiox (approved):
🚀

PR #385, opened 2023-03-17
feat: Update high congestion error message

cyaiox:
This PR updates the high congestion error message.

image


src/modules/toast/toasts/meta-transactions.tsx
```diff
@@ -76,7 +76,21 @@ export function getHighCongestionErrorToast(): Omit<Toast, 'id'> {
     title: t('@dapps.toasts.meta_transactions.high_congestion_error.title'),
     body: (
       <MetaTransactionError
-        text={t('@dapps.toasts.meta_transactions.high_congestion_error.body')}
+        text={
+          <span>
+            {t('@dapps.toasts.meta_transactions.high_congestion_error.body', {
+              a: (text: string) => (
+                <a
+                  href="https://polygonscan.com/gastracker"
```
fzavalia, line 84: Shouldn't this discriminate also mumbai in case we are using testnet? might be unnecessary but doesn't feel right to use polygonscan for testnet

src/modules/toast/toasts/meta-transactions.tsx
```diff
@@ -76,7 +76,21 @@ export function getHighCongestionErrorToast(): Omit<Toast, 'id'> {
     title: t('@dapps.toasts.meta_transactions.high_congestion_error.title'),
     body: (
       <MetaTransactionError
-        text={t('@dapps.toasts.meta_transactions.high_congestion_error.body')}
+        text={
+          <span>
+            {t('@dapps.toasts.meta_transactions.high_congestion_error.body', {
+              a: (text: string) => (
```
fzavalia, line 82: Will this replace <a> tags inside the translation?

PR #393, opened 2023-03-28
feat: Not enough allowance

fzavalia:


src/modules/authorization/reducer.ts
```diff
@@ -74,9 +84,26 @@ export function authorizationReducer(
...
     }
+    case FETCH_AUTHORIZATION_SUCCESS: {
+      const { authorization } = action.payload
+      const data = state.data.map(stateAuthorization => {
+        if (authorization && areEqual(stateAuthorization, authorization)) {
+          return authorization
+        } else {
+          return stateAuthorization
+        }
+      })
```
LautaroPetaccio, line 95: I'm not sure I understand this section of code, is it trying to replace the authorization with the new one?
What do you think about doing something like this?
```suggestion
      const data = state.data.filter(stateAuthorization => !areEqual(stateAuthorization, authorization))
      data.push(authorization)
```

src/modules/authorization/reducer.ts
```diff
@@ -89,10 +116,23 @@ export function authorizationReducer(
...
           const { authorization } = transaction.payload
+          let authorizationExists = false
+          const data = state.data.map(stateAuthorization => {
+            if (areEqual(stateAuthorization, authorization)) {
+              authorizationExists = true
+              return authorization
+            } else {
+              return stateAuthorization
+            }
+          })
```
LautaroPetaccio, line 127: This section of code seems to be doing the same as above, what do you think about replacing it with a filter and a push?

src/modules/authorization/utils.ts
```diff
@@ -16,6 +16,37 @@ export function hasAuthorization(
   )
 }
 
+export function hasAuthorizationAndEnoughAllowance(
```
LautaroPetaccio, line 19: I'm not sure where it's used, but, if we're storing the authorizations in the store, shouldn't this be a selector?

src/modules/authorization/utils.ts
```diff
@@ -16,6 +16,37 @@ export function hasAuthorization(
...
+  authorizations: Authorization[],
+  authorizationToFind: Authorization,
+  allowance: string
```
LautaroPetaccio, line 22: What if we change the name of this variable to:
```suggestion
  requiredAllowance: string
```

src/modules/authorization/sagas.ts
```diff
@@ -122,6 +133,64 @@ export function createAuthorizationSaga() {
...
+    const { authorization } = action.payload
+
+    try {
+      if (!isValidType(authorization.type)) {
+        throw new Error(`Invalid authorization type ${authorization.type}`)
+      }
+
+      const { chainId } = authorization
+
+      const provider: Provider = yield call(() => getNetworkProvider(chainId))
```
LautaroPetaccio, line 148: Using an arrow function will cause the effect to be untestable, would you mind fixing it?
```suggestion
      const provider: Provider = yield call(getNetworkProvider, chainId)
```

src/modules/authorization/sagas.ts
```diff
@@ -179,4 +248,30 @@ export function createAuthorizationSaga() {
   }
 }
 
+function getERC20ContractInstance(
```
LautaroPetaccio, line 251: Is there a reason why we're instantiating the contract instead of using the objects from the decentraland-transactions repo?

LautaroPetaccio (changes requested):
Great job! I've left some comments to be reviewed, but it should work perfectly

juanmahidalgo (approved):
LGTM!

src/modules/authorization/sagas.ts
```diff
@@ -179,4 +248,30 @@ export function createAuthorizationSaga() {
...
+  return new ethers.Contract(
+    authorization.contractAddress,
+    [
+      'function allowance(address owner, address spender) view returns (uint256)'
+    ],
+    provider
+  )
+}
+
+function getERC721ContractInstance(
```
cyaiox, line 264: // TODO: Use decentraland-transactions

LautaroPetaccio (approved):

PR #396, opened 2023-04-03
fix: Dropped watched transaction error

LautaroPetaccio:
This PR does the following:
- Updates the fetch transaction mechanism to use the `from` address in the request instead of the wallet's address which might differ after connecting with other wallet.
- Updates the `handleReplaceTransactionRequest` handler to get the network provided on each loop to prevent request to a Metamask provider that is not connected at the time.


cazala (approved):
Looking great!

PR #408, opened 2023-04-19
feat(WC2): Use WalletConnectV2 depending on FF

fzavalia:


fzavalia:
Closing because this solution is invalid

PR #411, opened 2023-04-24
feat: Allow links inside profile

kevinszuchet:


src/containers/Profile/Profile.tsx
```diff
@@ -1,42 +1,48 @@
...
-  static defaultProps = {
-    inline: true
-  }
+const Profile = function<T extends React.ElementType>(props: Props<T>) {
+  const { address, debounce, inline = true } = props
+  const profiles = useSelector(getProfiles)
+  const dispatch = useDispatch()
 
-  timeout: NodeJS.Timeout | null = null
+  let timeout: NodeJS.Timeout | null = null
```
LautaroPetaccio, line 13: As we're using a functional component, the contents of timeout won't be there after a render re-trigger. To make it persistent, we can use `useRef`:
```suggestion
  let timeout: number | null = useRef(null)
```

LautaroPetaccio (changes requested):
Great job 👏

src/containers/Profile/Profile.tsx
```diff
@@ -1,42 +1,47 @@
...
 
-  render() {
-    return <BaseProfile {...this.props} />
-  }
+    return () => {
+      if (timeoutRef.current) {
+        clearTimeout(timeoutRef.current)
+      }
+    }
+  }, [address, timeoutRef])
```
LautaroPetaccio, line 42: If I'm not wrong, only address needs to be taken into account to trigger the `useEffect` here. If not, the effect will be re-triggered multiple times as the `timeoutRef` will change every time the `useEffect` is executed and there's no avatar and there's a debounce.
```suggestion
  }, [address])
```

src/containers/Profile/Profile.tsx
```diff
@@ -1,42 +1,47 @@
...
-  static defaultProps = {
-    inline: true
-  }
+const Profile = function<T extends React.ElementType>(props: Props<T>) {
+  const { address, debounce, inline = true } = props
+  const profiles = useSelector(getProfiles)
+  const dispatch = useDispatch()
 
-  timeout: NodeJS.Timeout | null = null
+  const timeoutRef = useRef<NodeJS.Timeout>()
```
LautaroPetaccio, line 13: Is it possible by any chance to use the type that the `setTimeout` function has in the dom or is it bothering at compile time?
```suggestion
  const timeoutRef = useRef<number>()
```

LautaroPetaccio (approved):

PR #418, opened 2023-05-01
refactor: Clean state when fails enabling/connecting wallet without dispatching disconnectWallet

kevinszuchet:
Related to https://github.com/decentraland/marketplace/issues/1619

### DIFF
--- a/src/modules/wallet/sagas.spec.ts
+++ b/src/modules/wallet/sagas.spec.ts
@@ -1,24 +1,111 @@
 import { expectSaga } from 'redux-saga-test-plan'
 import * as matchers from 'redux-saga-test-plan/matchers'
-import { call } from 'redux-saga/effects'
+import { call, race, take } from 'redux-saga/effects'
 import { ChainId } from '@dcl/schemas/dist/dapps/chain-id'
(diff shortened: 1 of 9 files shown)


src/modules/wallet/sagas.spec.ts
```diff
@@ -1,25 +1,113 @@
...
+        return expectSaga(walletSaga)
+          .provide([
+            [
+              race({
+                success: take(FETCH_WALLET_SUCCESS),
+                failure: take(FETCH_WALLET_FAILURE)
+              }),
+              { success: null, failure: { payload: { error } } }
+            ]
+          ])
```
LautaroPetaccio, line 47: What if we mock, in the provide section, the `put(fetchWalletRequest())` so the action doesn't get handled?
The same could be applied to the other tests of the saga that put the same action.

LautaroPetaccio (commented):
Looks great ✅ I've just left a small comment on a test.

LautaroPetaccio (approved):
Send it 🚀

cazala (approved):
LGTM ✅

PR #421, opened 2023-05-03
feat: Add non-retryable API errors mechanism for the BaseClient

kevinszuchet:
This PR refactors the retrial mechanism in the `BaseClient`.

If the API will throw the same error every time the client makes the exactly same request, the `BaseClient` implementation should not try to repeat the failed request.

### DIFF
--- a/src/lib/BaseClient.spec.ts
+++ b/src/lib/BaseClient.spec.ts
@@ -40,7 +40,7 @@ beforeEach(() => {
   jest.clearAllMocks()
 })
 
-describe('when the request fails', () => {
+describe('when the request fails with a server error', () => {
   beforeEach(() => {
     nock(urlTest)
       .get('/test')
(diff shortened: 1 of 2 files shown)


flobarreto (approved):

src/lib/BaseClient.ts
```diff
@@ -12,12 +13,15 @@ export type BaseClientConfig = {
...
   retryDelay?: number
+  /** The status numbers that would not be retried */
+  nonRetryableStatuses?: number[]
 }
 
 export abstract class BaseClient {
   private readonly baseUrl: string
   private readonly retries: number
   private readonly retryDelay: number
+  private readonly nonRetryableStatuses: number[]
```
cyaiox, line 24: should this logic be added to the `BaseAPI.ts` too?

LautaroPetaccio: I think we shouldn't because it would be interesting to start deprecating the `BaseAPI` class.
The idea behind the `BaseClient` is to be a way to move away from Axios to the authenticated fetch from the `decentraland-crypto-fetch`. This fetch contains the logic to perform authenticated requests and uses the fetch interface for making requests, making it more compatible with other components that we're building.
We can move features that are in the `BaseAPI` class that we don't have here though.

src/lib/BaseClient.spec.ts
```diff
@@ -153,6 +153,35 @@ describe('when the request fails', () => {
   })
 })
 
+describe('when the request fails with a client error non-retryable error', () => {
```
LautaroPetaccio, line 156:
```suggestion
describe('when the request fails with a client non-retryable error', () => {
```

LautaroPetaccio (approved):
Looks good! I've left a small comment on what seems to be a typo. Just dismiss it if it's not.

PR #425, opened 2023-05-04
feat: add authorization flow saga

meelrossi:
Add new actions and saga to handle the whole authorization flow
- Grant/Revoke token
- Wait for transaction to be confirmed
- Fetch authorizations
- Check that the authorization was done correctly

### DIFF
--- a/src/modules/authorization/actions.spec.ts
+++ b/src/modules/authorization/actions.spec.ts
@@ -8,10 +8,15 @@ import {
 } from './actions'
 import { Authorization, AuthorizationAction } from './types'
 
+let authorization: Authorization
+
 describe('authorization flow actions', () => {
+  beforeEach(() => {
+    authorization = {} as Authorization
+  })
+
   describe('when calling authorizationFlowRequest action', () => {
     it('should return the correct action type and payload', () => {
-      const authorization = {} as Authorization
       const authorizationAction = AuthorizationAction.GRANT
       const allowance = '10'
       expect(
@@ -29,7 +34,6 @@ describe('authorization flow actions', () => {
 
   describe('when calling authorizationFlowSuccess action', () => {
     it('should return the correct action type and payload', () => {
-      const authorization = {} as Authorization
       expect(authorizationFlowSuccess(authorization)).toEqual({
         type: AUTHORIZATION_FLOW_SUCCESS,
         payload: {
@@ -41,7 +45,6 @@ describe('authorization flow actions', () => {
 
   describe('when calling authorizationFlowFailure action', () => {
     it('should return the correct action type and payload', () => {
-      const authorization = {} as Authorization
       const error = 'some error'
       expect(authorizationFlowFailure(authorization, error)).toEqual({
         type: AUTHORIZATION_FLOW_FAILURE,
--- a/src/modules/authorization/actions.spec.ts
+++ b/src/modules/authorization/actions.spec.ts
@@ -1,0 +1,55 @@
+import {
+  AUTHORIZATION_FLOW_FAILURE,
+  AUTHORIZATION_FLOW_REQUEST,
+  AUTHORIZATION_FLOW_SUCCESS,
+  authorizationFlowFailure,
+  authorizationFlowRequest,
+  authorizationFlowSuccess
+} from './actions'
+import { Authorization, AuthorizationAction } from './types'
+
+describe('authorization flow actions', () => {
+  describe('when calling authorizationFlowRequest action', () => {
+    it('should return the correct action type and payload', () => {
+      const authorization = {} as Authorization
+      const authorizationAction = AuthorizationAction.GRANT
+      const allowance = '10'
+      expect(
+        authorizationFlowRequest(authorization, authorizationAction, allowance)
+      ).toEqual({
+        type: AUTHORIZATION_FLOW_REQUEST,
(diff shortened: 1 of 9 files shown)


src/modules/authorization/actions.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+  AUTHORIZATION_FLOW_FAILURE,
+  AUTHORIZATION_FLOW_REQUEST,
+  AUTHORIZATION_FLOW_SUCCESS,
+  authorizationFlowFailure,
+  authorizationFlowRequest,
+  authorizationFlowSuccess
+} from './actions'
+import { Authorization, AuthorizationAction } from './types'
+
+describe('authorization flow actions', () => {
```
LautaroPetaccio, line 11: What do you think about making `authorization` global? It would remove the authorization definition from all the tests.
```suggestion
let authorization: Authorization

describe('authorization flow actions', () => {
  beforeEach(() => {
    authorization = {} as Authorization
  })
```

src/modules/authorization/reducer.spec.ts
```diff
@@ -81,3 +84,108 @@ describe.each([
...
+          loading: []
+        })
+      )
+    })
+
+    it('should remove error', () => {
+      const initialStateWithError = {
+        ...INITIAL_STATE,
+        authorizationFlowError: 'Some error'
+      }
```
LautaroPetaccio, line 144: What if we make the success action and the initial state global to the scope of the describe? By doing this, we could avoid repeating the initialization of the variable and initializing both resources in the describe.

```suggestion
    let successAction: AuthorizationSuccessAction
    let initialState: RootState
    beforeEach(() => {
      initialState = {
        ...INITIAL_STATE,
        loading: [successAction],
        authorizationFlowError: 'Some error'
      }
    })
    it('should remove action from loading array', () => {
      expect(
        authorizationReducer(initialStateWithLoading, successAction)
      ).toEqual(
        expect.objectContaining({
          loading: []
        })
      )
    })

    it('should remove error', () => {
```

src/modules/authorization/sagas.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+  contractAddress: '0x9c76ae45c36a4da3801a5ba387bbfa3c073ecae3',
+  authorizedAddress: '0x9c76ae45c36a4da3801a5ba387bbfa3c073ecae4',
+  contractName: ContractName.MANAToken,
+  chainId: ChainId.ETHEREUM_GOERLI
+}
+
+describe('handleAuthorizationFlowRequest', () => {
+  describe('when granting a token', () => {
+    describe('and authorization flow finishes successfully', () => {
+      it('should put success action', () => {
```
LautaroPetaccio, line 42: If I'm not wrong, as we want to refer to a specific action (the success one) we must use `the`.
```suggestion
      it('should put the success action', () => {
```
If that's correct, the same would apply to the other describes mentioning an specific action.

src/modules/authorization/sagas.ts
```diff
@@ -143,6 +170,7 @@ export function createAuthorizationSaga() {
 
   function* handleRevokeTokenRequest(action: RevokeTokenRequestAction) {
     const { authorization } = action.payload
+    console.log('HOLAAA')
```
LautaroPetaccio, line 173: Oops, it seems a console log got away.

src/modules/authorization/sagas.ts
```diff
@@ -155,6 +183,91 @@ export function createAuthorizationSaga() {
...
+          throw new Error(AuthorizationError.INSUFFICIENT_ALLOWANCE)
+        }
+
+        if (
+          authorization.type === AuthorizationType.APPROVAL &&
+          !hasAuthorization(authorizations, authorization)
+        ) {
+          throw new Error(AuthorizationError.GRANT_FAILED)
+        }
+      }
```
LautaroPetaccio, line 263: Would you mind adding tests for these parts of the code? I might be wrong, but I don't think they're covered.

LautaroPetaccio (changes requested):
Looks great! I've just left some comments about the tests

src/modules/authorization/sagas.ts
```diff
@@ -155,6 +183,91 @@ export function createAuthorizationSaga() {
...
+        success: RevokeTokenSuccessAction | GrantTokenSuccessAction | undefined
+        failure: RevokeTokenFailureAction | RevokeTokenFailureAction | undefined
+      } = yield race({
+        success: take(TOKEN_SUCCESS),
+        failure: take(TOKEN_FAILURE)
+      })
+      if (failure) {
+        throw new Error(failure.payload.error)
+      }
+      const txHash = getTransactionFromAction(success!).hash
```
cyaiox, line 213: could you please modify the non-null assertion?

```suggestion
      const txHash = success && getTransactionFromAction(success).hash
```

meelrossi: Done! I added `success as TokenSuccessAction` because if i did the `success && ` then i would have to solve txHash being undefined also which is a case that can't happen as success or failure should have a value and if failure has it will throw an exception so when reaching the code success has to be different than undefined

cyaiox (approved):

LautaroPetaccio (approved):
Send it 🚀

PR #439, opened 2023-05-22
fix: add possiblity to close authorization modal

meelrossi:
This pr fixes several issues with authorization modal
- Add possiblity to close it from the wrapped component
- Add possibility to set custom authorization names and titles. (TODO: there will come a PR soon that improves all the translation handling from other dapps. So that all the specific translation for each modal can be set from the dapp directly. For now its just a custom logic to improve publish collection modal )
- Only show to decimals in modal
- Add `t_cond` function so that we can pass a default translation if the first one is not defined. The logic to see if its defined or not is that t('id') == 'id' as it is the default value taken by [react-intl ](https://formatjs.io/docs/react-intl/components/#message-formatting-fallbacks) if all previous formatters fail


src/containers/withAuthorizedAction/AuthorizationModal/AuthorizationModal.tsx
```diff
@@ -45,7 +44,7 @@ export function AuthorizationModal({
   const [shouldReauthorize, setShouldReauthorize] = useState(false)
 
   const requiredAllowanceAsEth = requiredAllowance
```
cyaiox, line 46:
```suggestion
  const requiredAllowanceAsEth = getPriceInMana(requiredAllowance)
```

PR #455, opened 2023-07-04
feat!: Bump libraries to support Sepolia

lauti7:


src/lib/peer.ts
```diff
@@ -1,15 +1,16 @@
-import { LambdasClient } from 'dcl-catalyst-client/dist/LambdasClient'
+import { createLambdasClient, LambdasClient } from 'dcl-catalyst-client/dist/client/LambdasClient'
 import { Profile } from '../modules/profile/types'
 import { BaseAPI } from './api'
 import { FetchProfileOptions, ProfileEntity } from './types'
+import { createFetchComponent } from '@well-known-components/fetch-component'
```
meelrossi, line 5: can you move this import up?

src/lib/peer.ts
```diff
@@ -27,10 +28,8 @@ export class PeerAPI extends BaseAPI {
     }
 
     this.cache[address] = this.lambdasClient
-      .fetchProfiles([address.toLowerCase()], {
-        headers: { 'Cache-Control': 'max-age=0' }
-      })
-      .then(profiles => profiles[0])
+      .getAvatarsDetailsByPost({ ids: [address.toLowerCase()] })
```
meelrossi, line 31: is the cache control no needed anymore?

src/modules/profile/sagas.spec.ts
```diff
@@ -115,7 +115,7 @@ describe('when handling the action to set the profile avatar description', () =>
 
 describe('when handling the action to set the profile avatar alias', () => {
   const alias = 'anAlias'
-
+
```
meelrossi, line 118: can you remove this space 😬

src/lib/peer.ts
lauti7: I think that as the request is done by the POST method, the request is not cached by the browser, so no need of adding the cache control header. Besides that, the new function doesn't accept sending headers

fzavalia (approved):
Given that dependencies had major updates, release this as a major change. To do so, update the merge commit name to something like explained in semantic-release docs https://github.com/semantic-release/semantic-release#commit-message-format

fzavalia:
I've updated the PR name so it is deployed as a mayor 😄

PR #472, opened 2023-07-28
feat: added asset card

flobarreto:


src/containers/AssetCard/AssetCard.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { useMemo } from 'react'
+import { Item } from '@dcl/schemas'
+import { formatWeiToAssetCardEther, getCatalogCardInformation } from './utils'
+import {
+  AssetCard as AssetCardUi,
+  AssetCardProps as AssetCardUiProps
+} from 'decentraland-ui'
```
LautaroPetaccio, line 7: Would you mind changing this to import it from the dist directory? This is to prevent loading the whole UI when using the container.

PR #507, opened 2023-09-20
feat: Add RarityFilter

LautaroPetaccio:
This PR adds the RarityFilter component so it can be used in both the Profile site and the Marketplace.


src/components/RarityFilter/RarityFilter.tsx
```diff
@@ -0,0 +1,86 @@
...
+        )}
+      </div>
+    ),
+    [rarities, isMobileOrTablet, allRaritiesSelected]
+  )
+
+  return (
+    <Box
+      header={header}
+      className={'dui-rarity-filter' + (className ? ' ' + className : '')}
```
meelrossi, line 73: do we want to differenciate the components from ui and dapps?

meelrossi (approved):

PR #530, opened 2023-12-18
feat: update the profile sagas deployEntityWithoutNewFiles to use the identity

juanmahidalgo:


src/modules/profile/sagas.ts
```diff
@@ -134,15 +137,18 @@ export function createProfileSaga({
...
-
-    yield call(
-      [entities, 'deployEntityWithoutNewFiles'],
-      profileMetadata,
-      getHashesByKeyMap(newAvatar),
-      EntityType.PROFILE,
-      address,
-      address
-    )
+    const identity = getIdentity()
```
LautaroPetaccio, line 140: With the new code:
- If there's an identity, the `deployEntityWithoutNewFiles` is called, meaning that the entity is deployed.
- If there's no identity, the `deployEntityWithoutNewFiles` is not called and nothing happens.

In the older code, if the deployment failed, the  the update process failed as well, but this is not the case anymore. We should throw if there's no identity set, what do you think?

LautaroPetaccio (commented):
Great job! I've left a small comment to review.

LautaroPetaccio (approved):
🚀

PR #548, opened 2024-01-04
feat: Add Navbar2 container

cyaiox:
This PR adds a new Navbar2 container that implements the renewed user navbar.


src/containers/Navbar2/Navbar.container.tsx
```diff
@@ -0,0 +1,48 @@
+import { connect } from 'react-redux'
```
LautaroPetaccio, line 1: The directory and the component is called `Navbar2`, what do you think about renaming these files to `Navbar2`?

src/containers/Navbar2/Navbar.tsx
```diff
@@ -0,0 +1,144 @@
...
+      setTimeout(() => {
+        window.open(`${BASE_URL}/account`, '_blank', 'noopener')
+      }, 300)
+    },
+    [analytics]
+  )
+
+  const handleClickNavbarItem = useCallback(
+    (
+      _e: React.MouseEvent,
```
LautaroPetaccio, line 50: If any of the navbar item clicks imply a redirection or going to another page, if we don't wait for the tracking to be finished, it won't be tracked.

cyaiox: here is a timeOut defined when this `CB` is called: https://github.com/decentraland/ui/blob/master/src/components/Navbar2/SubMenuItem/SubMenuItem.tsx#L20-L33
```
const handleClick = useCallback(
    (event: React.MouseEvent<HTMLElement, MouseEvent>) => {
      event.preventDefault()
      onClickMenuOption &&
        onClickMenuOption(event, { eventTrackingName, url: href, isExternal })
      setTimeout(
        () => {
          window.open(href, isExternal ? '_blank' : '_self', 'noopener')
        },
        onClickMenuOption ? 300 : 0
      )
    },
    [onClickMenuOption, eventTrackingName, href, isExternal]
  )
```

LautaroPetaccio (approved):

PR #566, opened 2024-01-18
feat: Introduce FIAT Gateway

juanmahidalgo:
This PR
- Introduces the `FiatGateways` with `WERT` as the first supported type. 
- Moves the `identity` module from the marketplace to here so it's easier to use from the different dapps


src/modules/identity/utilts.ts
```diff
@@ -0,0 +1,18 @@
+import { AuthIdentity } from 'decentraland-crypto-fetch'
+import { localStorageGetIdentity } from '@dcl/single-sign-on-client'
+
+export function* getIdentityOrRedirect(
```
LautaroPetaccio, line 4: It seems that we're using the `getIdentityOrRedirect` from the sagas, should we keep this function as well?

src/modules/identity/actions.ts
```diff
@@ -0,0 +1,27 @@
+import { action } from 'typesafe-actions'
+import { AuthIdentity } from '@dcl/crypto'
+
+// Generate identity
+
+export const GENERATE_IDENTITY_REQUEST = '[Request] Generate Identity'
```
LautaroPetaccio, line 6: In theory, there's no need to keep this actions, as we won't be handling them anywhere after the refactor in the sites. Are you leaving them to be compatible with what's built today?

src/modules/gateway/sagas.ts
```diff
@@ -83,268 +100,341 @@ export function createGatewaySaga(config: ManaFiatGatewaySagasConfig) {
...
+            )
 
-      if (pendingManaPurchase.gateway === NetworkGatewayType.MOON_PAY) {
-        goToUrl = new MoonPay(config.moonPay).getTransactionReceiptUrl(
-          pendingManaPurchase.id
-        )
+            if (!identity) {
+              yield put(openFiatGatewayWidgetFailure(NO_IDENTITY_ERROR))
+              return
+            } else if (identity) {
```
LautaroPetaccio, line 123: WDYT about simplifying this?:
```suggestion
            if (!identity) {
              yield put(openFiatGatewayWidgetFailure(NO_IDENTITY_ERROR))
              return
            }
```

src/modules/gateway/sagas.ts
```diff
@@ -83,268 +100,341 @@ export function createGatewaySaga(config: ManaFiatGatewaySagasConfig) {
...
+                        // it's a success event
+                        onSuccess?.({ data: options, type: 'payment-status' })
+                      } else {
+                        onPending?.({ data: options, type: 'payment-status' })
+                      }
+                    }
+                  }
+                })
+
+                wertWidget.open()
```
LautaroPetaccio, line 171: Would you mind wrapping the this line with a `yield call`? If not, the code will try to instantiate the modal and that might be time consuming.
```suggestion
                yield call([wertWidget, ['open'])
```

LautaroPetaccio (approved):

src/modules/identity/utilts.ts
juanmahidalgo: removed!

src/modules/identity/actions.ts
juanmahidalgo: exactly, leaving it here so it's compatible with how it's used nowadays. If it's not used in the future, I'll remove it.

PR #567, opened 2024-01-18
feat: add the identity module

juanmahidalgo:
Moves the module defined [here](https://github.com/decentraland/marketplace/tree/master/webapp/src/modules/identity) to this repo


src/modules/identity/sagas.ts
```diff
@@ -0,0 +1,156 @@
...
+  GENERATE_IDENTITY_REQUEST,
+  GenerateIdentityRequestAction,
+  generateIdentityFailure,
+  generateIdentityRequest,
+  generateIdentitySuccess
+} from './actions'
+import { getConnectedProvider } from '../../lib/eth'
+import { Provider } from '../wallet/types'
+
+type IdentitySagaConfig = {
```
juanmahidalgo, line 30: This is new

PR #568, opened 2024-01-23
feat: add address field

meelrossi:
Add AddressField component with resolveAddress function


src/components/AddressField/AddressField.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import {
```
LautaroPetaccio, line 2: Would you mind importing this component directly from the distributable directory of `decentraland-ui`?

LautaroPetaccio (approved):

PR #583, opened 2024-02-22
feat: add timeout to switch network saga and show toast if it fails

juanmahidalgo:


src/modules/wallet/sagas.ts
```diff
@@ -199,8 +203,22 @@ function* handleSwitchNetworkRequest(action: SwitchNetworkRequestAction) {
...
+        timeout: delay(SWITCH_NETWORK_TIMEOUT) // 10 seconds timeout
+      })
+
+      if (timeout) {
+        yield put(
+          switchNetworkFailure(
+            chainId,
+            'Error switching network: Operation timed out'
+          )
+        )
```
LautaroPetaccio, line 217: If you want, you can avoid repeating the failure action by throwing:
```suggestion
      if (timeout) {
        throw new Error('Error switching network: Operation timed out')
```

LautaroPetaccio (approved):

src/modules/wallet/sagas.ts
juanmahidalgo: great idea! updated!

PR #588, opened 2024-03-05
feat: Add new cross chain support for transactions

LautaroPetaccio:


src/modules/transaction/utils.ts
```diff
@@ -98,13 +157,18 @@ export function buildTransactionWithFromPayload(
...
 }
 
 export function getTransactionHref(
-  { txHash, address, blockNumber }: TransactionHrefOptions,
+  { txHash, address, blockNumber, isCrossChain }: TransactionHrefOptions,
   network?: number
 ) {
+  if (isCrossChain) {
+    return `https://axelarscan.io/gmp/${txHash}`
+  }
```
juanmahidalgo, line 170: should we grab this from the AxelarProvider?

src/modules/transaction/types.ts
```diff
@@ -3,6 +3,7 @@ import {
   TransactionReceipt
 } from '@ethersproject/providers'
 import { ChainId } from '@dcl/schemas/dist/dapps/chain-id'
+import { AnyAction } from 'redux'
```
juanmahidalgo, line 6: can we move this import to the top?

src/modules/transaction/types.ts
```diff
@@ -105,3 +113,8 @@ export type Log = {
   name: string
   [key: string]: any
 }
+
+export type TransactionsConfig = {
+  squidUrl: string
```
juanmahidalgo, line 118: I know it's not that abstract right now to plug&plug another provider, but should we try to keep it generic and use something like `crossChainProviderURL`? I know we have `SQUID_URL` as the config env and Axelar as the provider name, but I'm just trying to not keep incorporating any of those names in here, wdyt?

src/modules/transaction/sagas.ts
```diff
@@ -46,13 +56,100 @@ import { isPending, buildActionRef } from './utils'
...
+    action: FetchCrossChainTransactionRequestAction
+  ) {
+    if (!config?.squidUrl) {
+      throw new Error('Squid URL not set')
+    }
+
+    const crossChainModule = import(
+      'decentraland-transactions/crossChain' as any
+    )
+    // const {
```
juanmahidalgo, line 85: whoops

src/modules/transaction/actions.ts
```diff
@@ -1,6 +1,7 @@
 import { AnyAction } from 'redux'
 import { action } from 'typesafe-actions'
 import { Transaction, AnyTransaction } from './types'
+import { ChainId } from '@dcl/schemas'
```
juanmahidalgo, line 4: can we move this one 1 line above?

src/modules/transaction/actions.ts
```diff
@@ -45,6 +46,58 @@ export type FetchTransactionFailureAction = ReturnType<
   typeof fetchTransactionFailure
 >
 
+// Fetch cross chain transaction
+
+export const FETCH_CROSS_CHAIN_TRANSACTION_REQUEST =
+  '[Request] Fetch Cross Transaction'
```
juanmahidalgo, line 52:
```suggestion
  '[Request] Fetch Cross Chain Transaction'
```

juanmahidalgo: right?

src/modules/transaction/utils.ts
```diff
@@ -224,3 +300,21 @@ export function* waitForTx(txHash: string) {
...
+  ...args: Parameters<(...args: any[]) => any>
+) =>
+  fork(function*() {
+    while (true) {
+      const action: FetchTransactionSuccessAction = yield take([
+        FETCH_TRANSACTION_SUCCESS,
+        FETCH_CROSS_CHAIN_TRANSACTION_SUCCESS
+      ])
+
+      if (action.payload.transaction.actionType === actionType) {
```
juanmahidalgo, line 316: not fully sure I undestand how this works, the `actionType` will be `FETCH_TRANSACTION_SUCCESS` or `FETCH_CROSS_CHAIN_TRANSACTION_SUCCESS` right? (the string values from those consts) but in the marketplace is called with `BUY_ITEM_CROSS_CHAIN_SUCCESS`. Will they match?

LautaroPetaccio: The transaction has the `actionType` which is the original action that triggered the fetch transaction action. We can track them this way and act upon it

src/modules/transaction/actions.ts
LautaroPetaccio: I've decided to remove the Cross Chain Transaction actions.

src/modules/transaction/actions.ts
LautaroPetaccio: Done, but removed the type.

src/modules/transaction/sagas.ts
LautaroPetaccio: Removed it!

src/modules/transaction/types.ts
LautaroPetaccio: I've changed it to `crossChainProviderUrl` to make it more generic.

src/modules/transaction/types.ts
LautaroPetaccio: I've moved it up!

src/modules/transaction/utils.ts
LautaroPetaccio: In order to avoid including the dependency here, I've changed the mechanism so we include in the transaction the cross chain provider, which is later used here to choose which URL to be written here.

src/modules/transaction/sagas.ts
```diff
@@ -42,17 +48,35 @@ import {
...
   yield takeEvery(WATCH_REVERTED_TRANSACTION, handleWatchRevertedTransaction)
   yield takeEvery(CONNECT_WALLET_SUCCESS, handleConnectWalletSuccess)
+
+  function* handleFetchTransactionRequest(
+    action: FetchTransactionRequestAction
+  ): any {
+    const isCrossChain = isTransactionActionCrossChain(action.payload.action)
+    if (isCrossChain) {
+      yield call(handleCrossChainTransactionRequest, action, config)
+    } else if (!isCrossChain) {
```
juanmahidalgo, line 76:
```suggestion
    } else  {
```
right?

src/modules/transaction/types.ts
```diff
@@ -78,22 +88,33 @@ export type Transaction = {
...
   payload?: any
   chainId: ChainId
 }
 
 export type TransactionPayload = {
-  [hash: string]: {
+  [key: string]: any,
+  [TRANSACTION_ACTION_FLAG]: {
     chainId: ChainId
+    toChainId: ChainId
```
juanmahidalgo, line 106: just a question, will this be a breaking change? this field is now required (just a comment so you know when releasing)

src/modules/transaction/utils.ts
```diff
@@ -22,60 +25,145 @@ import {
...
+}
+
+export function getTransactionPayloadFromAction(
+  action: ActionWithTransactionPayload
+): TransactionPayload['_watch_tx'] {
+  return action.payload[TRANSACTION_ACTION_FLAG]
+}
+
+export function isTransactionActionCrossChain(action: AnyAction): boolean {
+  return (
```
juanmahidalgo, line 78:
```suggestion
 const { chainId, toChainId } = getTransactionPayloadFromAction(action)
  return (isTransactionAction(action) && !!toChainId && chainId !== toChainId)
```
nit-pick: I think this looks a little bit cleaner and easy to read

juanmahidalgo (approved):

src/modules/transaction/sagas.ts
LautaroPetaccio: Removed it!

src/modules/transaction/types.ts
LautaroPetaccio: Made it optional because there surely are lots of transactions stored in the local storages without it.

src/modules/transaction/sagas.ts
```diff
@@ -82,7 +106,110 @@ export class FailedTransactionError extends Error {
...
+    typeof config.getCrossChainProvider
+  >> = yield config.getCrossChainProvider()
+  const crossChainProvider = new CrossChainProvider(
+    config.crossChainProviderUrl
+  )
+
+  let statusResponse: StatusResponse | undefined
+  let txInState: Transaction
+  let squidNotFoundRetries: number =
+    config.crossChainProviderNotFoundRetries ?? 10
```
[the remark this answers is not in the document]
LautaroPetaccio: Changed it to a constant with 120 retries.

src/modules/transaction/utils.ts
```diff
@@ -160,11 +264,23 @@ export function* waitForTx(txHash: string) {
       success,
       failure
     }: {
-      success: FetchTransactionSuccessAction | undefined
-      failure: FetchTransactionFailureAction | undefined
+      success:
+        | FetchTransactionSuccessAction
+        // | FetchCrossChainTransactionSuccessAction
```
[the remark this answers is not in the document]
LautaroPetaccio: Removed it!

PR #589, opened 2024-03-08
fix: Make gateways optional

fzavalia:
For the builder I only need wert, so having to define transak and moon pay is unnecessary.
This PR makes it possible to define only the desired gateways.

### DIFF
--- a/src/modules/gateway/sagas.ts
+++ b/src/modules/gateway/sagas.ts
@@ -132,14 +132,15 @@ export function createGatewaySaga(config: GatewaySagasConfig) {
               commodity,
               commodity_amount,
               sc_address,
-              sc_input_data
+              sc_input_data,
+              network
             } = data
             if (commodity && commodity_amount && sc_address && sc_input_data) {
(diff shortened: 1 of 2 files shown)


src/modules/gateway/sagas.ts
```diff
@@ -127,14 +132,15 @@ export function createGatewaySaga(config: GatewaySagasConfig) {
...
+              sc_input_data,
+              network
             } = data
             if (commodity && commodity_amount && sc_address && sc_input_data) {
               const dataToSign: WertMessage = {
                 address: wallet.address,
                 commodity,
                 commodity_amount,
-                network: isDev ? 'sepolia' : 'ethereum', // will be wallet.network
+                network: network ? network : isDev ? 'sepolia' : 'ethereum', // will be wallet.network
```
fzavalia, line 143: I will be needing mumbai and matic for collection publication fees

juanmahidalgo (approved):
LGTM

PR #594, opened 2024-03-18
feat: useNotifications hook

lauti7:


src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,109 @@
...
+      const interval = setInterval(() => {
+        fetchNotificationsState()
+      }, NOTIFICATIONS_QUERY_INTERVAL)
+
+      return () => {
+        clearInterval(interval)
+      }
+    } else {
+      return () => {}
+    }
```
braianj, line 92: do we need this? in case we do, I'll do it like this

```suggestion
    }
    return () => {}

```

src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,109 @@
...
+    if (identity) return new NotificationsAPI({ identity })
+    return null
+  }, [identity])
+
+  const handleOnBegin = () => {
+    setOnboardingDone()
+    setNotificationsState(prevState => ({ ...prevState, isOnboarding: false }))
+  }
+
+  const handleNotificationsOpen = async () => {
```
braianj, line 33: we should use a `useCallback` here

PR #619, opened 2024-05-10
feat: update base client with 204 response code

braianj:


src/lib/BaseClient.ts
```diff
@@ -57,7 +57,11 @@ export abstract class BaseClient {
         let parsedResponse: any
         try {
           response = await this.rawFetch(path, init)
-          parsedResponse = await response.json()
+          if(response.status === 204) {
+            parsedResponse = {ok: true, data: null}
+          } else {
+            parsedResponse = await response.json()
```
LautaroPetaccio, line 63: If you want the `BaseClient` to handle requests without responses, you can do that, generically, independently of the status code of the response, by checking if the body of the response is `null`. In those cases, you can set `parsedResponse` to `{ ok: true, data: null }` to let the flow continue and eventually return `null`.

src/lib/BaseClient.spec.ts
```diff
@@ -233,6 +233,29 @@ describe('when the request is successful', () => {
       expect(nock.isDone()).toBeTruthy()
     })
   })
+
+  describe("and the response return without content", () => {
```
LautaroPetaccio, line 237:
```suggestion
  describe("and the response has no body", () => {
```

src/lib/BaseClient.spec.ts
```diff
@@ -233,6 +233,29 @@ describe('when the request is successful', () => {
...
+
+      nock(urlTest)
+        .get('/test')
+        .reply(204)
+        .defaultReplyHeaders({
+          'Access-Control-Allow-Origin': '*'
+        })
+    })
+
+    it('should resolve with the response', async () => {
```
LautaroPetaccio, line 254:
```suggestion
    it('should resolve as null', async () => {
```

src/lib/BaseClient.ts
braianj: good catch

LautaroPetaccio (approved):

PR #625, opened 2024-06-04
feat: add new page tracking hook

meelrossi:
This PR is part of a series of PRs to stop using connected-react-router library. This PR starts sending the page event outside redux by using a hook when location changes

This is a breaking change as from now on we need to call `usePageTracking()` hooks in all our dapps

### DIFF
--- a/README.md
+++ b/README.md
@@ -849,21 +849,18 @@ export function* rootSaga() {
 }
 ```
 
-This uses by default the `'@@router/LOCATION_CHANGE'` action type to track page changes. If you need to use a different action type, you can do the following:
+### Page tracking
+
+In order to track all page change you will need to use analytics `page` function. There is already an exported hook you can use the will be triggered everytime a location change in the app
+
+Note: It is important that this hook is triggered in any component inside the router provider.
 
 ```ts
-import { all } from 'redux-saga/effects'
-import { createAnalyticsSaga } from 'decentraland-dapps/dist/modules/analytics/sagas'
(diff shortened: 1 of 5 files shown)


src/hooks/usePageTracking.ts
```diff
@@ -0,0 +1,15 @@
...
+import { useLocation } from 'react-router'
+import { getAnalytics } from '../modules/analytics'
+
+export function usePageTracking() {
+  const location = useLocation()
+
+  useEffect(() => {
+    const analytics = getAnalytics()
+    if (analytics) {
+      console.log({ location })
```
LautaroPetaccio, line 11: Would you mind removing the console log here?

README.md
```diff
@@ -849,21 +849,18 @@ export function* rootSaga() {
 }
 ```
 
-This uses by default the `'@@router/LOCATION_CHANGE'` action type to track page changes. If you need to use a different action type, you can do the following:
+### Page tracking
 
-```ts
-import { all } from 'redux-saga/effects'
-import { createAnalyticsSaga } from 'decentraland-dapps/dist/modules/analytics/sagas'
+In order to track all page change you will need to use analytics `page` function. There is already an exported hook you can use the will be triggered everytime a location change in the app
```
LautaroPetaccio, line 854:
```suggestion
In order to track all page change you will need to use the analytics `page` function. There is already an exported hook you can use the will be triggered everytime a location change in the app
```

LautaroPetaccio (approved):
Great job! Keep in mind that a commit with the BREAKING CHANGE message is needed to make it a breaking change.

PR #650, opened 2024-11-15
feat: remove marketplaceServerURL url from transakconfig

juanmahidalgo:


juanmahidalgo:
merging since it was already approved here: https://github.com/decentraland/decentraland-dapps/pull/646

PR #679, opened 2025-02-11
feat: avoid tracking if bot

braianj:


src/modules/analytics/utils.ts
```diff
@@ -64,6 +65,13 @@ export function isTrackable(action: AnyAction) {
 }
 
 export function getAnalytics() {
+  const userAgent = window.navigator.userAgent
+
+  const isBot = isbot(userAgent)
+  if (isBot) {
```
LautaroPetaccio, line 71: If we return null if the userAgent belong to a bot, we'll need to change how we access the analytics on all of our dApps, as they all assume that the analytics object is always returned.
The alternative to this is to return a mocked object that follows segment's write interface, that is, that has the functions `setAnonymousId`, `identify`, `track`, `page`, `trackLink` and `trackForm`

LautaroPetaccio (changes requested):

LautaroPetaccio (approved):

PR #685, opened 2025-02-17
feat: update lambda profile snapshot hash regex

juanmahidalgo:
# Switch Profile Fetching from Catalyst to Content Server

## Summary
Updated the profile fetching mechanism in the profile saga when deploying a new profile to use the Content Server (via `getProfileEntity`) instead of directly querying the Catalyst peer API (by `fetchProfile`). This change aligns with our move towards using the Content Server as the primary source for entity data and so we have the content hashes for the snapshots (the catalyst `fetchProfile` now returns the S3 URLs to the snapshots, we can't use them directly in the new deploy)

## Technical Changes
- Modified profile fetching logic in `profile/sagas.ts` to use `getProfileEntity` from the entities library
- Removed direct Catalyst API calls for profile fetching


kevinszuchet (approved):

LautaroPetaccio (approved):
LGTM 👍

PR #689, opened 2025-02-24
feat: start using cdn to get campaign content

braianj:


src/modules/campaign/ContentfulClient.spec.ts
```diff
@@ -1,87 +1,194 @@
...
+        await expect(
+          client.fetchEntry('space-id', 'env-id', 'entry-id')
+        ).rejects.toThrow('Failed to fetch entity data')
+      })
+    })
+  })
+
+  describe('fetchEntryAllLocales', () => {
+    it('should combine fields from all available locales', async () => {
+      const mockResponses = {
```
LautaroPetaccio, line 61: Would you mind building the context in a before each?

src/modules/campaign/ContentfulClient.spec.ts
```diff
@@ -1,87 +1,194 @@
...
-          mockToken
-        )
-
-        expect(fetch).toHaveBeenCalledWith(
-          `https://cdn.contentful.com/spaces/${mockSpace}/environments/${mockEnvironment}/entries/?sys.id=${mockId}&content_type=${mockContentType}&locale=*`,
-          {
-            headers: {
-              Authorization: `Bearer ${mockToken}`
+  describe('fetchAssetsFromEntryFields', () => {
+    it('should fetch all assets referenced in the entry fields', async () => {
```
LautaroPetaccio, line 118: Would you mind using the describe to explain each context?
```suggestion
   describe("when there are fields to get the assets from", () =>
    it('should get all assets referenced in the entry fields', async () => {
```

src/modules/campaign/ContentfulClient.spec.ts
```diff
@@ -1,87 +1,194 @@
...
+      expect(Object.keys(result)).toHaveLength(1)
     })
 
-    describe('when the request fails', () => {
-      beforeEach(() => {
-        global.fetch = jest.fn().mockResolvedValue({
-          ok: false
-        })
-      })
+    it('should return empty object when no assets are found', async () => {
```
LautaroPetaccio, line 146: Same as above:
```suggestion
   describe("when there are no entry fields to look assets from", () =>
    it('should return empty object', async () => {
```

src/modules/campaign/ContentfulClient.spec.ts
```diff
@@ -1,87 +1,194 @@
...
+              type: 'Link',
+              linkType: 'Asset',
+              id: marketplaceHomepageBannerAssets[0].sys.id
             }
           }
-        )
+        }
+      }
+
+      global.fetch = jest.fn().mockResolvedValue({
```
LautaroPetaccio, line 131: Would you mind defining the context in a `beforeEach`?

src/modules/campaign/ContentfulClient.spec.ts
```diff
@@ -1,87 +1,194 @@
...
+        enUS: {
+          fields: { title: { [ContentfulLocale.enUS]: 'English Title' } }
+        },
+        es: {
+          fields: { title: { [ContentfulLocale.es]: 'Título en Español' } }
+        },
+        zh: { fields: { title: { [ContentfulLocale.zh]: '中文标题' } } }
+      }
+
+      global.fetch = jest
```
LautaroPetaccio, line 71: What if we use nock here as well?

src/modules/campaign/ContentfulClient.spec.ts
```diff
@@ -1,87 +1,194 @@
...
+          json: () => Promise.resolve(mockResponses.zh)
+        })
+
+      const result = await client.fetchEntryAllLocales(
+        mockSpace,
+        mockEnvironment,
+        mockId
+      )
+
+      expect(result).toEqual({
```
LautaroPetaccio, line 92: We're generally missing the cases where the requests fail, would you mind adding tests for these cases?

src/modules/campaign/sagas.ts
```diff
@@ -27,62 +25,73 @@ export function* campaignSagas(
...
-            const bannerEntry = entries[linkedEntryId]
-            if (
-              bannerEntry &&
-              bannerEntry.sys.contentType.sys.id === BANNER_CONTENT_TYPE
-            ) {
-              acc[key] = {
-                ...bannerEntry.fields,
-                id: linkedEntryId
-              } as BannerFields & { id: string }
+      const banners = arrayOfFields
```
LautaroPetaccio, line 63: Would you mind explaining why are we using the `arrayOfFields` here? As far as I remember, we should only go through the fields of the admin entry to get the banner entries, right?

LautaroPetaccio (commented):
Great work! The PR looks great. There are some changes that we should do to the tests before merging though.

src/modules/campaign/sagas.ts
braianj: good catch

src/modules/campaign/sagas.ts
```diff
@@ -92,8 +97,10 @@ export function* campaignSagas(
       }) as LocalizedField<SysLink<'Entry'>> | undefined
 
       const campaignFields = campaignField?.[ContentfulLocale.enUS].sys.id
-        ? (entries[campaignField?.[ContentfulLocale.enUS].sys.id]
-            ?.fields as CampaignFields)
+        ? (Object.values(entries).find(
```
LautaroPetaccio, line 100: Shouldn't the entries be a record of <`entry.sys.id`, entry>? If that's the case, there's no need to go through all of the entries, right?

src/modules/campaign/ContentfulClient.spec.ts
```diff
@@ -1,87 +1,273 @@
...
+          mockEnvironment,
+          []
+        )
+        expect(result).toEqual({})
+      })
+    })
+  })
+
+  describe('fetchEntriesFromEntryFields', () => {
+    it('should fetch all entries referenced in the entry fields', async () => {
```
LautaroPetaccio, line 224: This test does not match the testing standard. Would you mind doing the same that you did with the other tests here? The context must be described for each `it` and a `beforeEach` must be used to define the context.

src/modules/campaign/ContentfulClient.spec.ts
```diff
@@ -1,87 +1,273 @@
...
+          client.fetchAsset(mockSpace, mockEnvironment, mockId)
+        ).rejects.toThrow('Failed to fetch asset data')
       })
     })
   })
+
+  describe('fetchAssetsFromEntryFields', () => {
+    describe('when there are fields to get the assets from', () => {
+      it('should get all assets referenced in the entry fields', async () => {
+        global.fetch = jest.fn().mockResolvedValue({
```
LautaroPetaccio, line 195: This test is using a global fetch, would you mind replacing it with nock so all tests are the same? The nock should be defined in a `beforeEach`

LautaroPetaccio (approved):

src/modules/campaign/ContentfulClient.spec.ts
braianj: good catch

src/modules/campaign/sagas.ts
braianj: good catch

PR #699, opened 2025-04-27
feat: use credits SSE endpoint

juanmahidalgo:
WIP

### DIFF
--- a/src/modules/credits/sagas.ts
+++ b/src/modules/credits/sagas.ts
@@ -5,8 +5,12 @@ import {
   delay,
   take,
   select,
-  race
+  race,
+  fork,
+  cancel,
+  cancelled
 } from 'redux-saga/effects'
+import { Task, eventChannel, EventChannel, END } from 'redux-saga'
 import { isErrorWithMessage } from '../../lib/error'
 import { CONNECT_WALLET_SUCCESS, ConnectWalletSuccessAction } from '../wallet'
 import {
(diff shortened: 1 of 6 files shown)


src/modules/credits/sagas.ts
```diff
@@ -91,4 +101,108 @@ export function* creditsSaga(options: { creditsClient: CreditsClient }) {
...
+    sseConnectionTask = yield fork(startSSEConnection, address)
+  }
+
+  let sseConnectionTask: Task | null = null
+  let currentSSEConnection: EventSource | null = null
+
+  function createCreditsEventChannel(address: string): EventChannel<any> {
+    return eventChannel((emit: (input: any) => void) => {
+      // Check if credits feature is enabled
+      const isCreditsEnabled = select(isCreditsFeatureEnabled, address)
```
kevinszuchet, line 130: Are we missing a yield here?

src/modules/credits/sagas.ts
```diff
@@ -91,4 +101,108 @@ export function* creditsSaga(options: { creditsClient: CreditsClient }) {
       attempts++
     }
   }
+
+  function* handleStartSSEConnection(
+    action: StartCreditsSSEAction
+  ): Generator<any, void, any> {
```
kevinszuchet, line 107: Are we ok with the `Generator<any, void, any>` response types?

src/modules/credits/sagas.ts
```diff
@@ -91,4 +110,116 @@ export function* creditsSaga(options: { creditsClient: CreditsClient }) {
...
+      if (!isCreditsEnabled) {
+        emit(END)
+        return () => {}
+      }
+
+      // Create SSE connection
+      const eventSource = options.creditsClient.createSSEConnection(
+        address,
+        (creditsData: CreditsResponse) => {
+          console.log('got creditsData', creditsData)
```
kevinszuchet, line 150: Should we remove this one?

src/modules/features/selectors.ts
```diff
@@ -163,6 +163,7 @@ export const isCreditsFeatureEnabled = (
   }
 
   const walletsAllowed = userWalletsVariant.payload.value
+    .trim()
```
kevinszuchet, line 166: Should we also trim the wallet just in case? I don't remember where you've done this already

kevinszuchet (approved):

PR #702, opened 2025-05-09
feat: Init Wert widget with the session id

kevinszuchet:
Use the session id besides the signature to init Wert's widget.


juanmahidalgo (approved):
🚀

PR #705, opened 2025-05-14
feat: add reconnect logic for credits sse + bump ui2

juanmahidalgo:


src/modules/credits/CreditsClient.ts
```diff
@@ -22,29 +22,105 @@ export class CreditsClient extends BaseClient {
...
+
+      eventSource = new EventSource(
+        `${this.url}/users/${address}/credits/stream`
+      )
+
+      // Copy all properties from the real EventSource to our wrapper
+      Object.defineProperties(
+        wrapper,
+        Object.getOwnPropertyDescriptors(eventSource)
+      )
```
LautaroPetaccio, line 59: This seems to be doing a shallow copy of eventSource, but it might not be doing what you're expecting, which is to copy the functionality from the EventSource, as it's prototyped and this new object is not linked to the EventSource object.

I've tried doing:
```suggestion
Object.getOwnPropertyDescriptors(new EventSource("https://credits.decentraland.org/users/0xaca5bc79b0cd51b726d2eadfc747f7ad4dfe7efb/credits/stream"))
```
And I got as a response a empty object.
You might want to instantiate it, cast it and modify it or take another path.

src/modules/credits/CreditsClient.ts
```diff
@@ -22,29 +22,105 @@ export class CreditsClient extends BaseClient {
...
+        // Forward the event if there's a listener on wrapper
+        if (wrapper.onopen) {
+          wrapper.onopen(event)
+        }
+      }
+
+      eventSource.onmessage = event => {
+        try {
+          const data = JSON.parse(event.data) as CreditsResponse
+          onMessage(data)
```
LautaroPetaccio, line 74: If we decide to handle the callbacks using the `onMessage` or the `onError` function parameters, what do you think about not returning the `EventSource`? We could return only a close function to kill the connection if needed.

src/modules/credits/CreditsClient.ts
```diff
@@ -22,29 +22,105 @@ export class CreditsClient extends BaseClient {
...
+          eventSource.close()
+        }
+
+        // Attempt to reconnect if under max retries
+        if (retryCount < MAX_RETRIES) {
+          retryCount++
+          console.log(
+            `Reconnecting SSE (attempt ${retryCount}/${MAX_RETRIES})...`
+          )
+          setTimeout(connect, RETRY_DELAY_MS)
```
LautaroPetaccio, line 103: If we expose the close function, we might want to store the `setTimeout` identifier and clear it when calling close, what do you think about it?

src/modules/credits/CreditsClient.ts
```diff
@@ -22,29 +22,105 @@ export class CreditsClient extends BaseClient {
...
+        console.error('SSE connection error:', error)
+
+        // Forward the event if there's a listener on wrapper
+        if (wrapper.onerror && eventSource) {
+          wrapper.onerror(error)
+        }
+
+        // Close the connection
+        if (eventSource) {
+          eventSource.close()
```
LautaroPetaccio, line 94: Should we set the eventSource as null here as well to not do a re-close when connect is called in the setTimeout?
```suggestion
          eventSource.close()
          eventSource = null
```

src/modules/credits/CreditsClient.ts
juanmahidalgo: sure, done

src/modules/credits/CreditsClient.ts
juanmahidalgo: good catch, approach changed

src/modules/credits/CreditsClient.ts
juanmahidalgo: sounds good, updated

LautaroPetaccio (approved):
🚀

PR #733, opened 2025-10-27
feat: Update Transak to generate the widget URL

LautaroPetaccio:
BREAKING: The Transak class interface changed.

This PR updates the Transak mechanism to instantiate and open a widget using the `Transak` class. In order to do that, the PR changes the following:
- The environment and api key are not longer used to instantiate the Transak SDK.
- The widget customization options are now part of the `openWidget` method.
- The `openWidget` method changed to request the Marketplace API for the widget URL.
- There's a new method in the Marketplace API client to retrieve the widget URL. The Marketplace API client is changed to properly use the BaseClient, which should be instantiated with the user's identity.

This PR is related to the marketplace-server one: https://github.com/decentraland/marketplace-server/pull/296

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "decentraland-dapps",
-  "version": "0.0.0-development",
+  "version": "22.0.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


juanmahidalgo (approved):
LGTM, just a couple of comments

PR #739, opened 2025-10-31
feat: add get identity id to auto sign in

braianj:
- Automatically creates and uses identityId when user signs in, removing the need to pass it as a prop
- Adds new `useSignInIdentity` hook to handle identityId creation logic
- Adds new `AuthAPI` module for authentication API interactions
- Uses generated identityId to create CDN links for navbar component

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,7 @@
         "decentraland-crypto-fetch": "^2.0.1",
         "decentraland-transactions": "^2.24.0",
         "decentraland-ui": "^6.22.0",
-        "decentraland-ui2": "^0.36.0",
+        "decentraland-ui2": "^0.38.0",
         "ethers": "^5.7.2",
         "events": "^3.3.0",
         "flat": "^5.0.2",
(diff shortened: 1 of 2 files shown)


src/modules/identityId.ts
```diff
@@ -0,0 +1,21 @@
+import { AuthAPI, AuthIdentityPayload } from './auth-api'
+import { AuthIdentity } from 'decentraland-crypto-fetch'
+
+export async function getIdentityId(identity: AuthIdentity): Promise<string> {
+  if (!identity) {
```
LautaroPetaccio: The function signature does not allow identity to be undefined, would you mind removing this check?

src/modules/identityId.ts
```diff
@@ -0,0 +1,21 @@
...
+    throw new Error('Identity is required')
+  }
+
+  if (!identity.authChain || !identity.ephemeralIdentity) {
```
LautaroPetaccio: The `authChain` and `ephemeralIdentity` are not undefined by how `AuthIdentity` is typed, would you mind removing these checks?

PR #761, opened 2025-12-23
fix: add fetch wrapper to prevent illegal invocation error in browser

braianj:


LautaroPetaccio (approved):

src/lib/fetcher.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * The @well-known-components/fetch-component package uses createFetchComponent()
+ * which internally passes fetch as a direct reference (globalThis.fetch).
+ * This causes fetch to lose its binding to the window object, resulting in
+ * "TypeError: Failed to execute 'fetch' on 'Window': Illegal invocation"
+ * when executed in the browser.
+ *
+ * This wrapper solves the problem by maintaining the correct context.
+ */
+export const fetcher = {
```
cyaiox: wdyt about renaming this wrapper to `const createFetchComponent = () => { fetch: ... }` in case we want to use it in other repos that also use `createFetchComponent`?

braianj: I can do that

PR #814, opened 2026-08-14
feat: allow loading analytics.js from a custom url

nachomazzara:
## What

Lets a dapp load `analytics.js` from its own URL instead of Segment's CDN:

```ts
const analyticsMiddleware = createAnalyticsMiddleware(SEGMENT_WRITE_KEY, {
  analyticsUrl: 'https://analytics.example.org/aPath/aBundle.min.js'
})
```

- `snippet.ts` is now Segment's snippet `5.2.0` (was `4.15.2`), written as `installAnalyticsSnippet`, still installed on import so nothing changes for the dapps that only import the middleware.
- `configureAnalyticsSnippet` sets the bundle URL and the `_cdn` `analytics.js` resolves its settings from, defaulting to the origin of the bundle URL. `createAnalyticsMiddleware` calls it with its new (optional) second argument before `load`.
- Both are exported from `modules/analytics`, and the middleware section of the README documents the option.

## Why

The snippet hardcoded `https://cdn.segment.com/analytics.js/v1/<write-key>/analytics.min.js` with no way to override it, and ad blockers drop that request, so those sessions send no events at all. Serving the same bundle from a first party proxy fixes it, but every dapp needs the URL to be configurable to do so.

Marketplace is the first consumer (its own workaround is https://github.com/decentraland/marketplace/pull/2679, which this replaces).

## Compatibility

- `createAnalyticsMiddleware(apiKey)` keeps working as is and keeps loading the bundle from Segment's CDN.
- The snippet upgrade brings the stubs of `screen` and `register`, the `data-global-segment-analytics-key` attribute, and the page context (`__t: 'bpc'`) that Segment attaches to the calls queued before the bundle loads, so they are attributed to the page they were made in.
- `installAnalyticsSnippet` and `configureAnalyticsSnippet` are no-ops when there is no `window`, the snippet used to throw on import outside the browser.

## Test

`src/modules/analytics/snippet.spec.ts` covers the queueing and the forwarding of calls once `analytics.js` is loaded, the default and the configured bundle URLs, the `_cdn` resolution, and the duplicated install.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/modules/analytics/snippet.ts
+++ b/src/modules/analytics/snippet.ts
@@ -30,6 +30,8 @@ const METHODS = [
   'ready',
   'alias',
   'debug',
+  // Dropped by Segment in 5.2.0, kept because `getAnonymousId` calls it
+  'user',
   'page',
   'screen',
   'once',
@@ -52,7 +54,12 @@ function getAnalyticsUrl(writeKey: string) {
 }
 
 function getOrigin(url: string) {
-  return new URL(url, window.location.href).origin
+  try {
+    return new URL(url, window.location.href).origin
+  } catch (_error) {
+    console.warn(`Analytics: could not resolve the origin of the analytics url "${url}"`)
+    return undefined
+  }
 }
 
 /**
--- a/src/modules/analytics/snippet.ts
+++ b/src/modules/analytics/snippet.ts
@@ -75,8 +75,13 @@ export function configureAnalyticsSnippet(newOptions: AnalyticsSnippetOptions = 
   const analytics = (window as unknown as { analytics?: SegmentSnippet }).analytics
 
   // analytics.js resolves the settings endpoint from here, it can't infer it from a proxied bundle path
-  if (analytics && options.cdnUrl) {
-    analytics._cdn = options.cdnUrl
+  if (analytics) {
+    if (options.cdnUrl) {
+      analytics._cdn = options.cdnUrl
+    } else {
+      // Reconfiguring back to Segment's CDN, leaving the previous one would resolve the settings from a stale origin
+      delete analytics._cdn
+    }
   }
 }
 
@@ -105,10 +110,12 @@ export function installAnalyticsSnippet() {
   analytics.methods = METHODS
   analytics.factory = (method: string) => {
     return (...args: unknown[]): unknown => {
-      // Once analytics.js is loaded the global is the real one, so calls kept from the snippet are forwarded to it
-      if (anyWindow.analytics?.initialized) {
-        const initializedMethod = anyWindow.analytics[method] as (...args: unknown[]) => unknown
-        return initializedMethod(...args)
+      // Once analytics.js is loaded the global is the real one, so calls kept from the snippet are forwarded to it,
+      // keeping it as the receiver because its methods rely on the analytics instance as `this`
+      const loadedAnalytics = anyWindow.analytics
+      if (loadedAnalytics?.initialized) {
+        const initializedMethod = loadedAnalytics[method] as (...args: unknown[]) => unknown
+        return initializedMethod.apply(loadedAnalytics, args)
       }
 
       if (METHODS_WITH_PAGE_CONTEXT.includes(method)) {
--- a/src/modules/analytics/snippet.ts
+++ b/src/modules/analytics/snippet.ts
@@ -1,73 +1,155 @@
-// @ts-nocheck
-;(function () {
-  // Create a queue, but don't obliterate an existing one!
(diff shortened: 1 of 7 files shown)


nachomazzara:
Addressed both findings in 2455fac.

**P1 — `user` stub.** Correct, `user` was in the 4.15.2 list and Segment dropped it in 5.2.0, so the method surface shrank. Added it back with a comment explaining why it deviates from the official snippet, plus a regression test.

One correction on the reported impact though: `getAnonymousId()` never worked before `analytics.js` loaded, on `master` either. The stub factory returns the queue array, so `analytics.user()` gave back an array and `.anonymousId()` threw `analytics.user(...).anonymousId is not a function`. Restoring the stub alone only changes which `TypeError` you get, so I also made `getAnonymousId` return `undefined` while the bundle is still loading, which matches what it already returns when analytics is unavailable. Covered by a test.

**P2 — trusted URLs.** `getOrigin` now warns and leaves `analytics.js` resolving the settings on its own instead of throwing on a malformed URL, and the README states both URLs are trusted build time values, never user input.

nachomazzara:
Both fixed in ccb3d83.

**P1 — receiver on forwarded calls.** Real bug, good catch. Segment's own snippet forwards with `window[i][e].apply(window[i], arguments)` and my port called the extracted method bare, so `this` was `undefined` for the real analytics methods that rely on it. Now applied with the loaded analytics as the receiver, with a test that fails without it (the mock returns `this`).

**P2 — stale `_cdn`.** Fixed too: `configureAnalyticsSnippet` deletes `_cdn` when there is no CDN URL, so reconfiguring back to Segment's CDN, or with a URL that can't be parsed, no longer leaves the settings resolving from the previous origin. Covered by two tests.

57 tests passing.

src/modules/analytics/snippet.ts
```diff
@@ -1,73 +1,162 @@
...
-  analytics.factory = function (method) {
-    return function () {
-      const args = Array.prototype.slice.call(arguments)
-      args.unshift(method)
-      analytics.push(args)
+  analytics.methods = METHODS
+  analytics.factory = (method: string) => {
+    return (...args: unknown[]): unknown => {
+      // Once analytics.js is loaded the global is the real one, so calls kept from the snippet are forwarded to it
+      if (anyWindow.analytics?.initialized) {
```
braianj: Could analytics[method] be undefined once analytics is initialized? For example, user was removed in Segment 5.2.0. Should we check that it’s a function before calling it?

nachomazzara:
Did a security pass over the whole diff, as an auditor would. One issue of my own found and fixed in ec6c077, plus the defense in depth suggested in the previous re-review.

**Untrusted URL reached `script.src` (would have been P1).** `configureAnalyticsSnippet` validated nothing: any string a dapp passed as `analyticsUrl` was assigned to the `src` of an injected script. The malformed URL case was worse than reported earlier, only the `_cdn` derivation was skipped while the raw value still got injected. Now both URLs are resolved and validated: anything that does not parse, or that is not `https:` unless it resolves to the dapp's own origin, is dropped with a warning and the bundle keeps loading from Segment's CDN. Fail closed, never fail open to an attacker controlled origin. Same-origin relative paths still work, which is what a proxy served by the dapp itself uses, and the dapp's own origin is allowed over http so local development keeps working.

**Write key interpolation.** The default URL interpolated the write key into the path unencoded. It cannot change the host, the authority is already terminated, but it is now `encodeURIComponent`'d so no key can reshape the path.

Rest of the surface reviewed, no findings:

- **Prototype pollution**: the only dynamic keys written to the global (`analytics[method]`) come from the module level `METHODS` constant, never from input.
- **Forwarding path**: the receiver fix landed in the previous commit, arguments are passed through untouched.
- **Page context**: `title`, `href`, `search` and `referrer` are attached to queued calls. This is the official 5.2.0 behavior and analytics.js already attaches the same context to every event once loaded, so no new category of data leaves the browser, it is only attributed to the page the call was made in rather than the one at replay time.
- **Secrets**: nothing hardcoded, the write key stays a runtime argument.
- **SSR**: both exported functions no-op without a `window`, the old snippet threw on import outside the browser.

Two operational notes that are out of scope for this PR but worth having somewhere: the bundle can't carry an SRI hash since Segment rotates it, so a CSP `script-src` allowlist is the practical control, and whoever owns the proxy host holds script execution rights on every dapp that points at it, so it deserves the same protection as the app's own origin.

Full suite green: 51 suites, 689 tests.

braianj (approved):
LGTM! 🎉

PR #827, opened 2026-08-31
feat: deliver analytics events to a first party ingestion host

braianj:
Ad blockers match Segment's hosts, so the migration to a first party proxy is only half done for the dapps that go through this package. Marketplace and Builder already load `analytics.js` and its settings from the proxy via `analyticsUrl` / `cdnUrl`, but every event still POSTs to `api.segment.io`, which the same filter lists block. `@dcl/hooks` already exposes the equivalent option for React apps (`AnalyticsProvider`'s `apiHost`), and dapps on this snippet had no way to pass it.

What changes:

- `AnalyticsSnippetOptions` gains `apiHost`, the host events are delivered to, without a protocol and shaped `host/basePath` (for example `api.example.org/v1`), because `analytics.js` prepends the protocol and appends the method path (`/t`, `/i`, `/p`).
- It is validated like `analyticsUrl` and `cdnUrl` are: it must resolve to an https URL, a value carrying a protocol is accepted and stripped rather than producing `https://https://host`, and an invalid or plain http value is dropped with a console warning so the dapp degrades to Segment's own ingestion instead of pointing the SDK at an untrusted host. Same normalization as `resolveApiHost` in `@dcl/hooks`.
- The snippet's `analytics.load` forwards it as `integrations['Segment.io'].apiHost`, merged into whatever load options the caller passed and only when a valid `apiHost` is configured.
- `createAnalyticsMiddleware` now passes those load options to `load`, which it was dropping. This matters when `analytics.js` is already on the page, since then `load` is its own and reads nothing from the snippet.
- New `getAnalyticsLoadOptions`, for dapps that call `analytics.load` themselves instead of going through the middleware.

No feature flag is read here. Each app evaluates its own flag and decides whether to pass the proxy options, so this PR is only the capability.

How to test it:

- `npm test` covers the new behaviour in `src/modules/analytics/snippet.spec.ts` and `src/modules/analytics/middleware.spec.ts`: a configured `apiHost` reaching `load`'s integrations, a protocol-carrying or trailing-slash value being normalized, an invalid or http value being dropped with a warning, and omitting it leaving `load` exactly as it is today.
- In a dapp, link this branch and pass `apiHost` to `createAnalyticsMiddleware` alongside `analyticsUrl`. The events in the network tab should POST to `https://<apiHost>/t` instead of `https://api.segment.io/v1/t`. Without `apiHost` they should keep going to `api.segment.io`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -895,7 +895,18 @@ const analyticsMiddleware = createAnalyticsMiddleware('SEGMENT WRITE KEY', {
 
 `analytics.js` fetches its settings from the origin of that URL. Pass `cdnUrl` as well when the proxy serves them from a different host.
 
-Both decide where a third party script is loaded from, so they are meant to be trusted URLs that come from the build configuration of the dapp, never from user input. Values that are not valid URLs, or that are not served over HTTPS unless they belong to the dapp's own origin, are ignored with a warning and the bundle keeps loading from Segment's CDN.
+The events themselves keep being delivered to Segment's ingestion endpoint, which the same filter lists drop. Pass `apiHost` to deliver them through the proxy too:
+
+```ts
+const analyticsMiddleware = createAnalyticsMiddleware('SEGMENT WRITE KEY', {
+  analyticsUrl: 'https://analytics.example.org/aPath/aBundle.min.js',
+  apiHost: 'api.example.org/v1'
+})
+```
+
+`apiHost` takes no protocol: `analytics.js` prepends it and appends the method path (`/t`, `/i`, `/p`), so the value is shaped `host/basePath`. A value that carries a protocol is accepted and stripped.
+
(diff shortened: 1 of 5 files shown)


braianj:
Addressed the P2s in eb6852d.

Added the three missing cases: a bare api host (`api.example.com`, no base path, asserting no trailing slash survives), load options unrelated to `integrations` (`obfuscate`, `initialPageview`) surviving the merge, and applying the merge to options that already went through it. All three already behaved correctly, so this is coverage rather than a fix. Also added the idempotency comment on `analytics.load`, spelling out that it is applied even though the middleware already applied it, because callers that install the snippet themselves never go through the middleware and merging the same host twice yields the same options.

Skipping the domain allowlist. The value comes from build configuration, not from user input, and it is already required to be an absolute https url. An allowlist would mainly make changing the proxy host harder without closing a path an attacker can actually reach. Happy to revisit if the host ever becomes runtime-configurable.

52 suites / 725 tests, build and prettier green.

